PR #6, opened 2017-11-02
Add clarification to point 3

rmoff:
Clarify from where `make clean all` needs to be run. It's there if you go the submodule route, but not explicit if you don't


rmoff:
@ybyzek is this OK to add?

ybyzek:
LGTM, thanks @rmoff

PR #15, opened 2017-12-28
Add 2-way SSL; Replicator; upgrade to CP 4.0

cotedm:
This is pull request that enables 2-way SSL for cp-demo for Kafka communication without endpoint verification and MD5 auth for ZooKeeper. There is one issue known to be solved where Control Center shows orange bars for consumption that will be addressed. I wanted to get the pull request started in case there are other reviews on documentation or style that need to be addressed that could make a larger impact on the structure.


Makefile
```diff
@@ -5,7 +5,7 @@ export PLUGINPATH:="connect-plugins"
 # fix for ubuntu sid sometimes linking /bin/sh to /bin/dash
 SHELL:=/bin/bash
 
-all: install irc transform
+all: install irc transform gencerts
```
ybyzek: This. Is. Awesome.

README.md
```diff
@@ -24,6 +25,8 @@ The use case is a streaming pipeline built around live edits to real Wikipedia p
 
 ![image](images/drawing.png)
 
+In this demo we have chosen to enable SSL for encryption and authentication for Kafka and SASL for Kafka communications with ZooKeeper. This provides example of how a production cluster would be configured in a secured environment.
```
ybyzek: "a production cluster" --> "components in the Confluent Platform" (to avoid people thinking that this Docker setup is for production)

scripts_ksql/setup.sh
```diff
@@ -7,12 +7,12 @@ fi
 
 DOCKER_MEMORY=$(docker system info | grep Memory | grep -o "[0-9\.]\+")
 if (( $(echo "$DOCKER_MEMORY 7.0" | awk '{print ($1 < $2)}') )); then
```
ybyzek: This needs to change to reflect 12GB (if we stick with 12GB)

docker-compose.yml
```diff
@@ -36,33 +50,63 @@ services:
...
+      KAFKA_JMX_PORT: 9991
+      KAFKA_SSL_KEYSTORE_FILENAME: kafka.kafka2.keystore.jks
+      KAFKA_SSL_KEYSTORE_CREDENTIALS: kafka2_keystore_creds
+      KAFKA_SSL_KEY_CREDENTIALS: kafka2_sslkey_creds
+      KAFKA_SSL_TRUSTSTORE_FILENAME: kafka.kafka2.truststore.jks
+      KAFKA_SSL_TRUSTSTORE_CREDENTIALS: kafka2_truststore_creds
+      # enables 2-way authentication
+      KAFKA_SSL_CLIENT_AUTH: "required"
+#      KAFKA_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM: "HTTPS"
+      KAFKA_ZOOKEEPER_SET_ACL: "true"
```
ybyzek: Maybe add a comment why this configuration is needed?

scripts_ksql/setup.sh
cotedm: looks like we don't need 12GB now that we've fixed some other configs, I'll remove this

Makefile
cotedm: Yep, it's trivial to add it to whatever new setup script comes into the fray. Happy to do so if this doesn't get merged before that happens.

cotedm:
latest commit adds in endpoint verification via HTTPS so now I think we have all of the SSL bells and whistles turned on

ybyzek:
Note: we removed ZK auth for now

PR #20, opened 2018-01-11
Update kafka-connect-irc connector; include jars, remove Makefile; update README and setup.sh scripts

cjmatta:
Updated to the latest version of kafka-connect-irc connector that implements a better `poll()` function to allow clean shutdown.


cjmatta:
LGTM

PR #21, opened 2018-01-12
Adding Authorizer/ACLs and related playbook

ybyzek:


cotedm:
One optional suggestion: I think adding an SSL only ACL example is valuable. The default authorizer includes the whole DN and this is a pain point for a lot of people trying to figure out how to operate around that string. Otherwise LGTM

ybyzek:
@cotedm : great suggestion, and I've added it in (replaced badclient)

PR #28, opened 2018-01-25
Remove reference to README from docs

joel-hamill:
Removed reference to README from docs/index

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -61,10 +61,10 @@ Follow along with the `Demo 1: Install + Run <https://www.youtube.com/watch?v=a4
 -  Docker Compose version 1.14.0 with Docker Compose file format 2.1
 -  Java version 1.8.0_92
 -  MacOS 10.12
--  GitHub
+-  git
 -  jq
 
-1. Clone the `demos GitHub repository <https://github.com/confluentinc/cp-demo>`__:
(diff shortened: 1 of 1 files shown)


joel-hamill:
ping @ybyzek

docs/index.rst
```diff
@@ -5,7 +5,7 @@ Monitoring Kafka streaming ETL deployments
 
 The Confluent Platform demo at `https://github.com/confluentinc/cp-demo <https://github.com/confluentinc/cp-demo>`__ shows users how to monitor Kafka streaming ETL deployments
 using `Confluent Control Center <http://docs.confluent.io/current/control-center/docs/index.html>`__.
-All the components in the |CP| have security enabled end-to-end. Follow along with the playbook in this README and watch the video tutorials.
+All the components in the |CP| have security enabled end-to-end.
```
ybyzek: Can we leave the line in but remove README?  For example: "Follow along with the playbook in this page and watch the video tutorials."

joel-hamill: in looking at this more, i think this first paragraph needs more work (e.g. it's not good information design to link people off to other pages in the first line). we want people reading about and using the demo from this page, not GitHub, so we should probably remove the link, at least in this prominent spot. i'll add to this PR.

PR #37, opened 2018-03-01
Adding check that Connect worker is online

SimonLeigh:
Adding check that Connect worker is online before submitting Connector configs to REST API.


ybyzek:
@SimonLeigh : thank you so much for the PR!  I tested it out and it works.  However, the log message it is checking for is misleading, i.e. "Finished starting connectors and tasks".  This message occurs several times, and as long as one appears it's ok to submit connectors, but it is confusing because no connectors have been submitted yet.  I would suggest instead checking for a message that is printed uniquely once, e.g. "Herder started".  Since I don't have the timing issue in my environment, could you test in your environment if checking on this log message works?

SimonLeigh:
@ybyzek : thanks for that suggestion - much better. It works and I've made the change.

ybyzek:
LGTM!

PR #38, opened 2018-03-21
improve stop.sh for deleting volumes

spahrson:
if 'docker ls' returns no volumes (cause you already stopped and cleaned everything) you cannot start cp-demo again, because start.sh is using stop.sh which is going to fail on xargs command with no (null) param.

### DIFF
--- a/scripts/stop.sh
+++ b/scripts/stop.sh
@@ -4,7 +4,6 @@ ps aux | grep ksql-server-start | grep -v "grep ksql" | awk '{print $2}' | xargs
 
 docker-compose down
 
-volumes=$(docker volume ls -q --filter="dangling=true")
-if [ -n "$volumes" ]; then
-	docker volumes rm "$volumes"
-fi
+for v in $(docker volume ls -q --filter="dangling=true"); do
+	docker volume rm "$v"
+done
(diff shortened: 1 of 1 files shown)


ybyzek:
@spahrson ,  in the scenario you describe where there are no volumes to remove:

* on your system, `xargs` fails unless you run it with `-r` option
* on my system, `xargs` doesn’t have `-r` option so this fails with `xargs: illegal option -- r`

Behavior of xargs varies from system to system.  BSD versions of xargs run with -r by default. GNU version needs it.

So this PR needs to be improved to take into account the system type and whether `-r` is needed and/or supported.

scripts/stop.sh
```diff
@@ -3,4 +3,8 @@
 ps aux | grep ksql-server-start | grep -v "grep ksql" | awk '{print $2}' | xargs kill -15
 
 docker-compose down
-docker volume ls -q --filter dangling=true | xargs docker volume rm
+
+volumes=$(docker volume ls -q --filter="dangling=true")
+if [ -n "$volumes" ]; then
+	docker volumes rm "$volumes"
```
ybyzek: * Command is `docker volume`, not `docker volumes`
* `docker volume` returns a list, so it needs to iterate through each item in the list

Consider changing code to:

```
volumes=$(docker volume ls -q --filter="dangling=true")
if [ -n "$volumes" ]; then
  docker volume ls -q --filter dangling=true | xargs docker volume rm
fi
```

ybyzek:
LGTM, thanks for your contribution @spahrson !

-Yeva

PR #43, opened 2018-04-19
Upgrade to CP 4.1

ybyzek:


cotedm:
LGTM, only comment I have is that I think that with 4.1 we can now also update this area as well https://github.com/confluentinc/cp-demo/blob/6c13dcbca3d942a47a1284944fd4812f62cdfe64/docker-compose.yml#L260

PR #47, opened 2018-04-24
Fix linux

cjmatta:
This will look up the name of the container instead of relying on a static name for `run_ksql.sh` as it seems like different versions of docker name the containers differently.


ybyzek:
No longer relevant, code has `docker-compose exec ksql-cli `

PR #50, opened 2018-05-23
DEVX-19: Add ZK authorization back in, now that Replicator has no ZK …

ybyzek:
…dependency


cotedm:
@ybyzek can you try the consumer throttling piece? I think this will fail with zk auth enabled because the throttle_consumer.sh script doesn't have a zk authenticated user defined. I'm not sure if you can use only a bootstrap server to define throttles or if you need zk specifically. If you can just go to the broker directly that'll be easier, otherwise, you'll need the zk jaas file for clients in KAFKA_OPTS

ybyzek:
@cotedm : excellent verification point.  The brokers are already configured with `KAFKA_OPTS` and the throttling script is run on the brokers -- I have just verified that it works.  I also augmented the docs to mention this, because I can see some users tripping over it.  Thank you!

PR #60, opened 2018-07-27
Use C3 URL that goes to 30 min view directly

rmoff:
Small tweak to URL to save the manual step of needing to change the time window

On Mac, will also open it directly after everything's started up

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -107,3 +107,6 @@ echo -e "\nWaiting for everything to stabilize, sleeping 30 seconds"
 sleep 30
 
 echo -e "\nDONE! Connect to Confluent Control Center at http://localhost:9021\n"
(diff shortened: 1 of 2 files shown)


scripts/start.sh
```diff
@@ -107,3 +107,6 @@ echo -e "\nWaiting for everything to stabilize, sleeping 30 seconds"
 sleep 30
 
 echo -e "\nDONE! Connect to Confluent Control Center at http://localhost:9021\n"
+if [[ $(uname -s) == "Darwin" ]]; then
+  open "http://localhost:9021/monitoring/system/brokers?latencyPercentile=95&rolling=30&view=RAW"
```
ybyzek: On my machine, this opens the page in Safari.  But it should use the Google Chrome application.

ybyzek:
Out-of-date PR, closing

PR #69, opened 2018-10-10
Added docs to address issue#68

MichaelHussey:
Modified docs to suggest explicitly specifying a branch


MichaelHussey:
Not needed - it turns out I managed to end up on _master_ instead of the current default _5.0.0-post_.
I must have explicitly set the branch at some stage and have subsequently forgotten.

PR #82, opened 2018-11-30
Fix the relative script references in start.sh

cjmatta:
There were a number of calls to the current directory instead of the fully qualified directory.


cjmatta:
I just noticed I branched this off of 5.2.0-SNAPSHOT, I'm going to close this PR in favor of one I just made off of 5.0.1-post

PR #88, opened 2019-01-25
DEVX-658, DEVX-657:

ybyzek:
1. DEVX-658: disable offset.timestamps.commit in Replicator so that wikipedia.parsed.replica does not show up in Control Center consumer lag (relevant when Replicator consumer lag shows in C3 in 5.1.1)
2. DEVX-657: do not overload `connect-replicator`, add comment to Docker compose

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -507,6 +507,8 @@ services:
     ports:
       - 8086:8086
 
+  # This container is just to transfer Replicator jars to the Connect worker
+  # It is not used as a Connect worker
   replicator:
     image: confluentinc/cp-enterprise-replicator:5.1.0
     hostname: replicator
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -507,6 +507,8 @@ services:
     ports:
       - 8086:8086
 
+  # This container is just to transfer Replicator jars to the Connect worker
+  # It is not used as a Connect worker
   replicator:
```
rmoff: If it's being used other than to extract JARs I would probably change the service/host/container name to something more indicative, i.e. not `replicator`.  

I'd also drop the environment clause entirely. 

For the `command:`, `sleep infinity` is my personal preference instead of `tail -f /dev/null`, it's more expressive/easier to grok at a glance what it's doing.

ybyzek: @rmoff excellent suggestions, I have addressed all 3 points

PR #89, opened 2019-01-31
Chrome "ERR_CERT_WEAK_SIGNATURE_ALGORITHM"

MarkC0x:
Added options to the certificate generation commands to increase security algorithm level.  This means that it should fix the Chrome "Your connection is not private" and "NET::ERR_CERT_WEAK_SIGNATURE_ALGORITHM" messages when trying to connect to C3 over SSL with Chrome.

### DIFF
--- a/scripts/security/certs-create.sh
+++ b/scripts/security/certs-create.sh
@@ -22,6 +22,8 @@ do
                                  -ext "SAN=dns:$i,dns:localhost" \
 				 -keystore kafka.$i.keystore.jks \
 				 -keyalg RSA \
+                                 -sigalg SHA256withRSA \
+                                 -keysize 2048 \				 
 				 -storepass confluent \
 				 -keypass confluent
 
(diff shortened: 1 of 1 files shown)


javabrett:
I ran this on a VM with a JDK8 from the current `5.3.0-post` branch.  The resulting certificate (I think this is the relevant one?) was:

```
openssl x509 -in control-center.certificate.pem -text

Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            b2:d8:0a:3e:5c:ce:39:c1
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=ca1.test.confluent.io, OU=TEST, O=CONFLUENT, L=PaloAlto, C=US
        Validity
            Not Before: Aug 19 00:03:28 2019 GMT
            Not After : Jan  3 00:03:28 2047 GMT
        Subject: C=US, ST=Ca, L=PaloAlto, O=CONFLUENT, OU=TEST, CN=control-center
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
                Modulus:

        X509v3 extensions:
            X509v3 Subject Alternative Name:
                DNS:control-center, DNS:localhost
    Signature Algorithm: sha256WithRSAEncryption
```
The `keytool` docs seem to suggest that if `RSA` is selected, it should select SHA256/2048 algorithms - looks like it might have done that above?  This might have improved with newer JDK keytool versions.  This was from `OpenJDK Runtime Environment (build 1.8.0_201-b09)`.

@MarkC0x wondering if you are seeing similar results on 5.3.0 that might make this change redundant?

MarkC0x:
That's interesting @javabrett -- your output does appear correct.

One other thing that was important was the signing of the certificate, which in my earlier case was showing up as SHA1, and that was the change made to Chrome to display the "Weak Signature" warning if it detected a SHA1 signed certificate.  The second command update in this PR was to force that Signature Algorithm to use SHA256.

However, looking again at your output, the Signature Algorithm also looks correct, so I am at a bit of a loss.  Maybe the "default_md" has also been updated sometime recently?

Either way, I am happy to close this out if it is no longer required to add these flags.

PR #97, opened 2019-03-26
DEVX-787: optimize start time

ybyzek:
Comparison of times for ```./scripts/start.sh``` to complete:

5.1.2-post: avg 4:03 (3:52, 4:16, 4:03)
PR w/ 3s gap wait: 2:59 (3:03, 3:02, 2:53)

Savings of 60+ seconds

Note:
Original PR w/ 10s gap wait: avg 3:17 (3:06, 3:24, 3:23)

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -40,8 +40,8 @@ MAX_WAIT=300
 CUR_WAIT=0
 echo "Waiting up to $MAX_WAIT seconds for Confluent Control Center to start"
 while [[ ! $(docker-compose logs control-center) =~ "Started NetworkTrafficServerConnector" ]]; do
-  sleep 10
-  CUR_WAIT=$(( CUR_WAIT+10 ))
+  sleep 3
+  CUR_WAIT=$(( CUR_WAIT+3 ))
   if [[ "$CUR_WAIT" -gt "$MAX_WAIT" ]]; then
     echo -e "\nERROR: The logs in control-center container do not show 'Started NetworkTrafficServerConnector' after $MAX_WAIT seconds. Please troubleshoot with 'docker-compose ps' and 'docker-compose logs'.\n"
     exit 1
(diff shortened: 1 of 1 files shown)


scripts/start.sh
```diff
@@ -107,8 +107,19 @@ echo -e "\nStart consumers for additional topics: WIKIPEDIANOBOT, EN_WIKIPEDIA_G
...
 ${DIR}/consumers/listen_EN_WIKIPEDIA_GT_1_COUNTS.sh
 
-echo -e "\nWaiting for KSQL queries to start, sleeping 50 seconds"
-sleep 50
+# Verify wikipedia.parsed topic is populated and schema is registered
+MAX_WAIT=50
+CUR_WAIT=0
+echo -e "\nWaiting up to $MAX_WAIT seconds for wikipedia.parsed topic to be populated"
+while [[ ! $(docker-compose exec schemaregistry curl -X GET --cert /etc/kafka/secrets/schemaregistry.certificate.pem --key /etc/kafka/secrets/schemaregistry.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt https://schemaregistry:8085/subjects) =~ "wikipedia.parsed-value" ]]; do
+  sleep 10
```
gnschenker: why 10 sec (instead of say 1 sec) if you really want to minimize startup time?

scripts/start.sh
```diff
@@ -107,8 +107,19 @@ echo -e "\nStart consumers for additional topics: WIKIPEDIANOBOT, EN_WIKIPEDIA_G
...
+  CUR_WAIT=$(( CUR_WAIT+10 ))
```
gnschenker: ditto

scripts/start.sh
ybyzek: Agree and makes sense.  I have reduced it to 3 seconds.  (1 second is a bit much, given that waiting for response from Docker takes ~1 second.)

PR #98, opened 2019-04-16
DEVX-16: improve cp-demo memory utilization

ybyzek:
This PR reduces overall load on the host system during the startup of the demo, but it does add time -- still under 4 minutes though.

Performance tests suggest that reducing the number of cpus (down from Docker default of 4 cpus) also reduces load, but it requires manual intervention (people may forget) or may not be desirable (people may choose not to do this).

Instead, this PR puts cpu limits on each container and applies configuration optimizations to Control Center


apurvam:
Thanks for the patch @ybyzek . I think this seems like the right direction if we want to programmatically solve the problem of the demo using unbound resources. The next step would be to figure out if it addresses @rmoff's issues on his laptop.

ybyzek:
@apurvam , agree!

Before merge:

1. @rmoff please confirm that this reduces load on your host machine.
2. Separately, I need to run through the whole `cp-demo` playbook to make sure all still works.

ybyzek:
@pdruley -- I haven't done any performance profiling with less than 8GB memory, so no comment at this time.  If memory is an issue, we can decouple from this PR and investigate separately.

@danphanley and @liko9 also validated the PR, so I'll go ahead and merge

PR #102, opened 2019-06-01
Add keytool to the list of commands to verify are installed

RothAndrew:
Add verification that the `keytool` command is installed. If it isn't installed the certificates won't get created, but docker-compose still runs (and fails).


ybyzek:
@RothAndrew -- thanks for the PR!  Can you kindly retarget `5.2.1-post` instead of `master`?

RothAndrew:
Sure no problem.

PR #103, opened 2019-06-02
Add keytool to the list of commands to verify are installed

RothAndrew:
Add verification that the `keytool` command is installed. If it isn't installed the certificates won't get created, but docker-compose still runs (and fails).


RothAndrew:
Oops that wasn't right...

PR #104, opened 2019-06-02
Add "keytool" to the list of applications that get verified

RothAndrew:
If `keytool` isn't installed, `docker-compose up` still runs, but the certificates never got created, so Kafka fails to start.


RothAndrew:
@ybzek here you go.

PR #108, opened 2019-07-17
DEVX-562: Disable provenance header explicitly (defaults to true <= 5.0)

rspurgeon:
We also want this to flow to `5.0.1-post`


ybyzek:
Note, the description is incorrect (I provided wrong information in the jira) -- we actually do NOT also want this in `5.0.1-post` branch since that is where the default changed to false

PR #116, opened 2019-08-08
EOF); causes warnings during start

DaveSlinn:
Having the EOF formatted as
```EOF);```
instead of
```
EOF
)
```
causes the following warning:
``warning: here-document at line 50 delimited by end-of-file (wanted `EOF')``.
By removing the semi-colon and adding a LF between the EOF and the ), the warning goes away.


ybyzek:
Thanks for the PR @DaveSlinn .  

Can you please retarget the PR to `5.3.0-post` branch instead of `master`?  That way the commit will apply for users running the demo against the current Confluent Platform release.

DaveSlinn:
Well, that didn't work. I will close and redo the PR.

PR #128, opened 2019-10-16
DEVX-1184: Catch when users try and use docker-compose up in cp-demo …

ybyzek:
…and fail fast

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,6 +29,7 @@ services:
     ports:
       - "9091:9091"
       - "29091:29091"
+    command: "bash -c 'if [ ! -f /etc/kafka/secrets/kafka.kafka1.keystore.jks ]; then echo \"ERROR: Did not find SSL certificates in /etc/kafka/secrets/ (did you remember to run ./scripts/start.sh instead of docker-compose up -d?)\" && exit 1 ; else /etc/confluent/docker/run ; fi'"
     environment:
       KAFKA_ZOOKEEPER_CONNECT: "zookeeper:2181"
       KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: SASL_SSL:SASL_SSL,SSL:SSL,SASL_SSL_HOST:SASL_SSL,PLAINTEXT:PLAINTEXT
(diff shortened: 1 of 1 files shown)


javabrett:
I know this is going to work - the only downside I see is that it overrides the image command, so if that ever changes then we have to catch that and maintain.

Maybe this could be done by adding/checking a mandatory Docker Compose variable binding (something set in the script) with a custom error message?

ybyzek:
@javabrett 

> I know this is going to work - the only downside I see is that it overrides the image command, so if that ever changes then we have to catch that and maintain.

It doesn't override it completely, it interjects a command before it runs the normal workflow.

> Maybe this could be done by adding/checking a mandatory Docker Compose variable binding (something set in the script) with a custom error message?

I'm not quite following.  Can you please elaborate on what this means, or provide an example, of what the check looks like, how it is called, etc.?

javabrett:
> It doesn't override it completely, it interjects a command before it runs the normal workflow.

My point was only that this portion:
```
; else /etc/confluent/docker/run
```
... is statically copied down from the image.  It's unlikely to change very often I agree.

ybyzek:
@javabrett Yeah, it's unlikely to change.

Can you please explain the alternate idea?

javabrett:
An alternate way to handle using Docker Compose env-var binding with mandatory-variables: https://github.com/confluentinc/cp-demo/compare/force-startup-via-start-sh

I exported this dummy variable and put it in ZooKeeper service def - probably doesn't need to be there but it is the most-depended-on service so I added it there.

Pros: Doesn't override container start command.  Possibly fails-faster since it should be checked at Docker Compose YAML parse-time.
Cons: Possibly more confusing.  Might not stop individual containers from being started incorrectly, but stopping ZK should be enough for now?

ybyzek:
@javabrett thanks for the example.  It works great (and I learned something cool, thank you!).  

However, with that solution all `docker-compose` commands result in errors:

```
$ docker-compose ps
ERROR: Missing mandatory value for "environment" option in service "zookeeper": ERROR:Detected direct 'docker-compose' command usage; instead run './scripts/start.sh' and './scripts/stop.sh'
```

And I think it adds more confusion and complicates things for users.  

We should prefer smoother/easier user experience (and acknowledge that we have the burden of updating the demo if `/etc/confluent/docker/run` changes).  So my vote is to leave the PR as is.  WDYT?

javabrett:
Agreed. The alternate approach would require lifecycle changes to set the variable in the current shell session, which is more complex.

Ship it!

PR #130, opened 2019-10-27
DEVX-1230: Replace all references of dest.kafka.replication.factor wi…

ybyzek:
…th confluent.topic.replication.factor


ybyzek:
> I cannot query the connect REST API for connector status using curl either outside or inside the connect container

@rspurgeon what specific command is failing?  Are you using HTTPS and associated required credentials, for example as shown here: https://github.com/confluentinc/cp-demo/blob/5.3.1-post/scripts/connectors/submit_elastic_sink_config.sh#L25

PR #133, opened 2019-11-15
DEVX-1032 Addresses warnings in key scripting

rspurgeon:
After a test it appears these are the only changes required to address the various warnings in the key generation scripting.  All services appear to function normally after this change.   Note the minor change in the CA subject... the state attribute is different for the CA request vs the keytool genkey.

### DIFF
--- a/scripts/security/certs-create.sh
+++ b/scripts/security/certs-create.sh
@@ -24,7 +24,7 @@ do
 				 -keyalg RSA \
 				 -storepass confluent \
 				 -keypass confluent \
-				 -deststoretype pkcs12
+				 -storetype pkcs12
 
 	# Create the certificate signing request (CSR)
 	keytool -keystore kafka.$i.keystore.jks -alias $i -certreq -file $i.csr -storepass confluent -keypass confluent -ext "SAN=dns:$i,dns:localhost"
--- a/scripts/security/certs-create.sh
+++ b/scripts/security/certs-create.sh
@@ -9,7 +9,7 @@
 rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12
 
 # Generate CA key
-openssl req -new -x509 -keyout snakeoil-ca-1.key -out snakeoil-ca-1.crt -days 365 -subj '/CN=ca1.test.confluent.io/OU=TEST/O=CONFLUENT/L=PaloAlto/S=Ca/C=US' -passin pass:confluent -passout pass:confluent
+openssl req -new -x509 -keyout snakeoil-ca-1.key -out snakeoil-ca-1.crt -days 365 -subj '/CN=ca1.test.confluent.io/OU=TEST/O=CONFLUENT/L=PaloAlto/ST=Ca/C=US' -passin pass:confluent -passout pass:confluent
 
 for i in kafka1 kafka2 client schemaregistry restproxy connect control-center
 do
@@ -23,7 +23,8 @@ do
 				 -keystore kafka.$i.keystore.jks \
 				 -keyalg RSA \
 				 -storepass confluent \
-				 -keypass confluent
+				 -keypass confluent \
+				 -deststoretype pkcs12
 
 	# Create the certificate signing request (CSR)
 	keytool -keystore kafka.$i.keystore.jks -alias $i -certreq -file $i.csr -storepass confluent -keypass confluent -ext "SAN=dns:$i,dns:localhost"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -504,6 +504,7 @@ services:
       - ksql-server
     volumes:
       - ./scripts/ksql/ksqlcommands:/tmp/ksqlcommands
+      - ./scripts/ksql/run_ksql_commands.sh:/tmp/run_ksql_commands.sh
     entrypoint: /bin/sh
     tty: true
   restproxy:
--- a/scripts/ksql/run_ksql.sh
+++ b/scripts/ksql/run_ksql.sh
@@ -1,8 +1,3 @@
 #!/bin/bash
 
-
-docker-compose exec ksql-cli  bash -c "ksql http://ksql-server:8088 <<EOF
-run script '/tmp/ksqlcommands';
-exit ;
-EOF
-"
+docker-compose exec ksql-cli bash -c "/tmp/run_ksql_commands.sh"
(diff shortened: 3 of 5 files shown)


scripts/security/certs-create.sh
```diff
@@ -9,7 +9,7 @@
 rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12
 
 # Generate CA key
-openssl req -new -x509 -keyout snakeoil-ca-1.key -out snakeoil-ca-1.crt -days 365 -subj '/CN=ca1.test.confluent.io/OU=TEST/O=CONFLUENT/L=PaloAlto/S=Ca/C=US' -passin pass:confluent -passout pass:confluent
+openssl req -new -x509 -keyout snakeoil-ca-1.key -out snakeoil-ca-1.crt -days 365 -subj '/CN=ca1.test.confluent.io/OU=TEST/O=CONFLUENT/L=PaloAlto/ST=Ca/C=US' -passin pass:confluent -passout pass:confluent
```
ybyzek: Do we understand when to use `S` or when to use `ST`?  What's the better practice?

rspurgeon:
> @rspurgeon when running this PR on my laptop, the KSQL command `select * from WIKIPEDIA;` fails (I've run it twice and it failed both times; I switched to 5.3.1-post without this change and it works).
> 
> If you haven't already, please validate that this change works end-to-end.

I ran end-to-end twice without noticing the error.  Can you please highlight where you see the error?

scripts/security/certs-create.sh
rspurgeon: Not really.  The documentation is weak.  `openssl req` produces this error warning if you use `S`: 

```
Subject Attribute S has no known NID, skipped
```

Change the attribute to `ST` and it goes away.  I assume that X509 has a different set of valid attributes or keytool / pkcs12 doesn't validate it.

ybyzek:
@rspurgeon the symptom was no records were returned.  Did `select * from WIKIPEDIA;` return records in your runs?

rspurgeon:
> @rspurgeon the symptom was no records were returned. Did `select * from WIKIPEDIA;` return records in your runs?

I see, it's a manual step from the playbook (?) which I didn't check. Will do now...

rspurgeon:
> @rspurgeon the symptom was no records were returned. Did `select * from WIKIPEDIA;` return records in your runs?

When i run this I _do_ see records from `SELECT * FROM WIKIPEDIA`.  I'm proceeding through the entire playbook again but on my most recent pass everything seemed to function

ybyzek:
@rspurgeon the issue I was encountering is something on my end -- it takes ~10 seconds for the first message to appear, during which C3 says no new messages.  This is a separate issue to follow up on, but it has nothing to do with this PR.  My apologies.

Before merging, let's make sure to get someone from @confluentinc/security to review/approve this PR.

rspurgeon:
> @rspurgeon the issue I was encountering is something on my end -- it takes ~10 seconds for the first message to appear

I've noticed this as well with C3

rspurgeon:
> @rspurgeon the issue I was encountering is something on my end -

In the meantime I've ran through the cp-demo playbook twice, focusing specifically on the Encryption & Auth sections: https://docs.confluent.io/current/tutorials/cp-demo/docs/index.html#encryption-authentication
and did not observe any issues

ybyzek:
Ping @jeffhuang26 for review

scripts/security/certs-create.sh
```diff
@@ -23,7 +23,8 @@ do
 				 -keystore kafka.$i.keystore.jks \
 				 -keyalg RSA \
 				 -storepass confluent \
-				 -keypass confluent
+				 -keypass confluent \
+				 -deststoretype pkcs12
```
jeffhuang26: I do not see option "-deststoretype" for -genkeypair command. Is it supposed to use "-storetype"?

ybyzek: Good catch @jeffhuang26 .  I wonder how this accidentally works.

@rspurgeon FWIW I tested with `storetype` instead of `deststoretype` and it appears to work as well:

```
cp-demo(DEVX-1032) ✗: git diff
diff --git a/scripts/security/certs-create.sh b/scripts/security/certs-create.sh
index e6e7381..bc6318b 100755
--- a/scripts/security/certs-create.sh
+++ b/scripts/security/certs-create.sh
@@ -24,7 +24,7 @@ do
                                 -keyalg RSA \
                                 -storepass confluent \
                                 -keypass confluent \
-                                -deststoretype pkcs12
+                                -storetype pkcs12
 
        # Create the certificate signing request (CSR)
        keytool -keystore kafka.$i.keystore.jks -alias $i -certreq -file $i.csr -storepass confluent -keypass confluent -ext "SAN=dns:$i,dns:localhost"
cp-demo(DEVX-1032) ✗: 
```

ybyzek:
@rspurgeon when running this PR in your environment, does it expose https://github.com/confluentinc/ksql/issues/3363 ?

scripts/security/certs-create.sh
rspurgeon: It appears they are aliased in the KeyTool source.  I copied from the referenced Jira.  Will update to use the flag referenced in the tool help.

rspurgeon:
> @rspurgeon when running this PR in your environment, does it expose [confluentinc/ksql#3363](https://github.com/confluentinc/ksql/issues/3363) ?

Investigating...

rspurgeon:
Ack, would be nice if GH supported moving a PR back to Draft status, but it does not ATM

big-andy-coates:
@ybyzek this PR has no relation to the bug in KSQL server that is causing the cp-demo to fail.  See: https://github.com/confluentinc/ksql/issues/3363#issuecomment-557499226

Please feel free to unblock this PR.

ybyzek:
@big-andy-coates based on experiences from two users (me and @rspurgeon ) the change in this PR changes timing _somehow_ (we don't know how) and increases the probability that the KSQL issue will surface.  In my case, it surfaces each and every run.  So they are related.

We cannot merge this PR until one of two things happens:
1. confluentinc/ksql#3363 is fixed for 5.4.0
2. cp-demo implements a workaround

So this PR is blocked until we determine which of the two options above is implemented.

ybyzek:
confluentinc/ksql#3363 is merged into 5.4.x (thanks @big-andy-coates!)

@rspurgeon can you please:

1. retarget this PR to 5.4.x branch
2. run one more validation test with 5.4.x as the base

PR #135, opened 2019-11-18
change zookeeper version to tie to CP and add license replication config

sdanduConf:
modified zookeeper version so it is the same as CP version instead of latest (i don't know if we guarantee that versions are compatible with newer Zookeepers in case that happens), but am okay if we decide we don't want that

modified parameter to deal with _confluent-licenses topic and only having 2 brokers in this demo.

Will update kafka-connect to server-connect in another PR (currently a blocking bug)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -152,7 +152,7 @@ services:
                   -Djavax.net.ssl.keyStore=/etc/kafka/secrets/kafka.kafka2.keystore.jks
                   -Djavax.net.ssl.keyStorePassword=confluent
   connect:
-    image: confluentinc/cp-kafka-connect:5.4.0-SNAPSHOT
+    image: confluentinc/cp-server-connect:5.4.0-SNAPSHOT
     container_name: connect
     cpus: 0.6
     restart: always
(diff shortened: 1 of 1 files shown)


sdanduConf:
> @sdanduConf thanks for the PR!
> 
> 1. Have you validated that cp-demo works with this change?
> 2. Do we need corresponding `EMIT CHANGES` added to https://github.com/confluentinc/cp-demo/blob/5.4.x/scripts/ksql/ksqlcommands

1. I've validated that cp-demo does indeed work
2) We don't need it for the ksqlcommands in that file, though if there are select queries elsewhere (in documentation, etc..), they might need updates.

sdanduConf:
@ybyzek also pushed the update from cp-kafka-connect -> cp-server-connect after testing

sdanduConf:
> @sdanduConf thanks for the PR. I ran the demo on latest 5.4 Docker image and it passes smoke test (I have not run through the whole playbook). I approve. Please let me know if there are more changes coming or if the PR is ready for merge.

No other changes planned - merging this PR.

PR #138, opened 2019-11-23
DEVX-1214: Ability to configure SSL truststore and keystore for SR Clients

ybyzek:


ybyzek:
Status: C3 still doesn't work due to MMA-5986 , but other services/clients appear to work.

PR #143, opened 2019-12-09
DEVX-1032 refactor ksql script handling into mapped files

rspurgeon:
This was my solution for running the ksql loading portion of the demo.  Added a few addl scripts which moved most of the processing into the docker container, which I hope makes things a little cleaner.

I targted DEVX-1032 first to isolate this change in a new PR, and then it will all be merged into 5.4.x after

### DIFF
--- a/scripts/ksql/run_ksql_commands.sh
+++ b/scripts/ksql/run_ksql_commands.sh
@@ -3,5 +3,5 @@
 input=/tmp/ksqlcommands
 while IFS= read -r line
 do
-  /tmp/run_ksql_cmd.sh "$line"
+  echo -e "$line\nexit" | ksql http://ksql-server:8088
 done < "$input"
--- a/scripts/ksql/run_ksql_commands.sh
+++ b/scripts/ksql/run_ksql_commands.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+input=/tmp/ksqlcommands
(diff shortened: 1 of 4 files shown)


scripts/ksql/run_ksql_commands.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+input=/tmp/ksqlcommands
+while IFS= read -r line
+do
+  /tmp/run_ksql_cmd.sh "$line"
```
ybyzek: @rspurgeon why require `run_ksql_cmd.sh`?  Why not just put here the content of `run_ksql_cmd.sh`?

rspurgeon: > @rspurgeon why require `run_ksql_cmd.sh`? Why not just put here the content of `run_ksql_cmd.sh`?

IMO the small files are clearer to read, and the single function "run_ksql_cmd" is potentially useful independently.  I could do a function in the script also, but this works.  If you want it smashed up into a single line i'll change it.  LMK

ybyzek: IMO it's better to mash it to avoid following a trail of files.   Thanks!

rspurgeon: Functional programmers might not agree ;-). I'll push a change

PR #146, opened 2019-12-16
ST-2811: Making the connection check to wikimedia irc work on rhel

elismaga:
On RHEL images the nc command output is formatted differently and does not contain the word we currently search for in order to determine success. I did test that this command will return a non-zero error code if it fails, so there is no need to parse the output. This will continue to work on Debian and RHEL images.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -74,10 +74,10 @@ while [[ ! $(docker-compose logs connect) =~ "Herder started" ]]; do
   fi
 done
 
-if [[ ! $(docker-compose exec connect timeout 3 nc -zv irc.wikimedia.org 6667) =~ "open" ]]; then
+docker-compose exec connect timeout 3 nc -zv irc.wikimedia.org 6667 || {
   echo -e "\nERROR: irc.wikimedia.org 6667 is unreachable. Please ensure connectivity before proceeding or try setting 'irc.server.port' to 8001 in scripts/connectors/submit_wikipedia_irc_config.sh\n"
   exit 1
-fi
+}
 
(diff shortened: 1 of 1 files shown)


scripts/start.sh
```diff
@@ -74,7 +74,9 @@ while [[ ! $(docker-compose logs connect) =~ "Herder started" ]]; do
   fi
 done
 
-if [[ ! $(docker-compose exec connect timeout 3 nc -zv irc.wikimedia.org 6667) =~ "open" ]]; then
+docker-compose exec connect timeout 3 nc -zv irc.wikimedia.org 6667
+
+if [[ $? != 0 ]]; then
   echo -e "\nERROR: irc.wikimedia.org 6667 is unreachable. Please ensure connectivity before proceeding or try setting 'irc.server.port' to 8001 in scripts/connectors/submit_wikipedia_irc_config.sh\n"
   exit 1
 fi
```
rspurgeon: Thanks @elismaga , for your consideration you could consider using the OR operator here to more explicitly tie the return code to the failure case.  It's slightly more explicit and will prevent the possibility that someone accidentally sneaks a command in front of the conditional and failure block.

```suggestion-5
docker-compose exec connect timeout 3 nc -zv irc.wikimedia.org 6667 || {
  echo -e "\nERROR: irc.wikimedia.org 6667 is unreachable. Please ensure connectivity before proceeding or try setting 'irc.server.port' to 8001 in scripts/connectors/submit_wikipedia_irc_config.sh\n"
  exit 1
}
```

elismaga: @rspurgeon Thanks for the suggestion.

PR #147, opened 2019-12-20
DEVX-1415: Write script that validates REST Proxy & SR

ybyzek:


scripts/app/validate_rest_proxy.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Register a new Avro schema for topic 'users'
+docker-compose exec schemaregistry curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --cert /etc/kafka/secrets/schemaregistry.certificate.pem --key /etc/kafka/secrets/schemaregistry.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt --data '{ "schema": "[ { \"type\":\"record\", \"name\":\"user\", \"fields\": [ {\"name\":\"userid\",\"type\":\"long\"}, {\"name\":\"username\",\"type\":\"string\"} ]} ]" }' https://schemaregistry:8085/subjects/users-value/versions
+
+# Get the Avro schema id
+schemaid=$(docker-compose exec schemaregistry curl -X GET --cert /etc/kafka/secrets/schemaregistry.certificate.pem --key /etc/kafka/secrets/schemaregistry.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt https://schemaregistry:8085/subjects/users-value/versions/1 | jq '.id')
```
rspurgeon: For your consideration, the above two lines could be combined into a single exec command with:

```
schemaid=$(docker-compose exec schemaregistry curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --cert /etc/kafka/secrets/schemaregistry.certificate.pem --key /etc/kafka/secrets/schemaregistry.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt --data '{ "schema": "[ { \"type\":\"record\", \"name\":\"user\", \"fields\": [ {\"name\":\"userid\",\"type\":\"long\"}, {\"name\":\"username\",\"type\":\"string\"} ]} ]" }' https://schemaregistry:8085/subjects/users-value/versions | jq '.id')
```

ybyzek: @rspurgeon thanks for the suggestion, but I'll leave it as two separate steps for consistency with how the steps are presented in the cp-demo docs.

ybyzek:
@rspurgeon 

1. The last line of the script deletes the consumer instance.  As a result, this script can be run more than once.  I have validated that running in multiple times in a row works.

```
cp-demo(DEVX-1415): ./scripts/app/validate_rest_proxy.sh 
{"id":7}{"offsets":[{"partition":0,"offset":0,"error_code":null,"error":null}],"key_schema_id":null,"value_schema_id":7}{"instance_id":"my_consumer_instance","base_uri":"https://restproxy:8086/consumers/my_avro_consumer/instances/my_consumer_instance"}[]
PASS: Output [{"topic":"users","key":null,"value":{"userid":1,"username":"Bunny Smith"},"partition":0,"offset":0}] matches expected output                           cp-demo(DEVX-1415): ./scripts/app/validate_rest_proxy.sh
{"id":7}{"offsets":[{"partition":1,"offset":0,"error_code":null,"error":null}],"key_schema_id":null,"value_schema_id":7}{"instance_id":"my_consumer_instance","base_uri":"https://restproxy:8086/consumers/my_avro_consumer/instances/my_consumer_instance"}[]
PASS: Output [{"topic":"users","key":null,"value":{"userid":1,"username":"Bunny Smith"},"partition":1,"offset":0}] matches expected output                           cp-demo(DEVX-1415): ./scripts/app/validate_rest_proxy.sh
{"id":7}{"offsets":[{"partition":0,"offset":1,"error_code":null,"error":null}],"key_schema_id":null,"value_schema_id":7}{"instance_id":"my_consumer_instance","base_uri":"https://restproxy:8086/consumers/my_avro_consumer/instances/my_consumer_instance"}[]
PASS: Output [{"topic":"users","key":null,"value":{"userid":1,"username":"Bunny Smith"},"partition":0,"offset":1}] matches expected output  
```

2. If a user registers the same exact schema outside of the script and then runs the script, it doesn't generate a new version of the schema, it uses the same version.

Generally, this script isn't being referenced anywhere from the docs, so it's a bit hidden and one has to look for it to find it.  As you alluded to, it's really as an out-of-band validation when someone changes cp-demo and wants an easy smoke test to ensure Schema Registry and REST Proxy still work.

rspurgeon:
>     The last line of the script deletes the consumer instance.  As a result, this script can be run more than once.  I have validated that running in multiple times in a row works.

doh!  I was running the script steps one by one outside the full script and didn't see this last line, my bad.

>     1. If a user registers the same exact schema outside of the script and then runs the script, it doesn't generate a new version of the schema, it uses the same version.

I was thinking of a scenario where a user might register a schema that results in a different version than this one and might result in multiple versions, so version 1 might be different than what is expected in this script.

PR #157, opened 2020-01-30
Master should be 5.5.x-latest not 5.5.0-SNAPSHOT

andrewegel:
Before we can get the tool team's tooling to automate these updates, we need to adhere to the set up standard, which is to use the `major.minor.x-latest` convention we're doing in the `major.minor.x` branches - I noticed that master is not following this convention right now, and also had a bad version in zookeeper.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "2.2"
 services:
   zookeeper:
-    image: confluentinc/cp-zookeeper:5.4.0-SNAPSHOT
+    image: confluentinc/cp-zookeeper:5.5.x-latest
     restart: always
     hostname: zookeeper
     container_name: zookeeper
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -184,7 +184,7 @@ services:
       CONNECT_PLUGIN_PATH: "/usr/share/java,/connect-plugins"
       CONNECT_LOG4J_ROOT_LOGLEVEL: INFO
       CONNECT_LOG4J_LOGGERS: org.reflections=ERROR
-      CLASSPATH: /usr/share/java/monitoring-interceptors/monitoring-interceptors-5.5.0-SNAPSHOT.jar
+      CLASSPATH: /usr/share/java/monitoring-interceptors/monitoring-interceptors-5.5.x-latest.jar
```
rspurgeon: For JARs, shouldn't we follow the SNAPSHOT pattern?

andrewegel: You're right, this was me doing a global sed, and it picked this up. Will fix.

andrewegel: fixed in next commit.

PR #164, opened 2020-03-06
DEVX-1270: RBAC

ybyzek:
Folks, when doing a review, please all expand `docs/index.html` and `docker-compose.yml` to see full diffs


daveshook:
I ran the demo using our latest ubuntu-18-04 classroom image running in VirtualBox. I am not sure if the issue mentioned for Linux still applies but just in case, I added `vm.max_map_count=262144` to `/etc/sysctl.conf`.

The first time I ran `./scripts/start.sh`, the `kafka-client` container timed out waiting for the broker so the topics didn't get created which caused everything that depended upon one of the topics to also fail. When I ran `./scripts/start.sh` a second time, the topics were created as were the streams and table. The rest of the demo behaved as expected.

One thing I am not sure about is the security. In the **Run Demo**, item 4 says to log in as `superuser` and password `superuser`. I am not sure of the context for this instruction. I opened C3 from my ubuntu VM without any prompt for user and password. I also opened C3 from my physical machine with the same result.

One last item to mention. The same item 4 mentions you may also log in as other users and `other users` is associated with an invalid url, `https://github.com/confluentinc/cp-demo/tree/5.4.1-post/scripts//security/ldap_users`.

daveshook:
I ran the demo using our latest ubuntu-18-04 classroom image running in VirtualBox. I am not sure if the issue mentioned for Linux still applies but just in case, I added `vm.max_map_count=262144` to `/etc/sysctl.conf`. I initially didn't checkout the `DEVX-1270` branch but fairly quickly figured out I needed to do so.

- [ ] In the **Run Demo** section, step 4  you may also log in as other users and `other users` is associated with an invalid url, `https://github.com/confluentinc/cp-demo/tree/5.4.1-post/scripts//security/ldap_users`.

- [ ] In the **Topics** section, step 6 says to examine the `wikipedia.parsed.count-by-channel` topic and then step 7  continues examining the `wikipedia.parsed` topic which was being examined in the first part of this section.

- [ ] In the **Topics** section, step 7, the screenshot looks very different from what I see.

- [ ] In the **Topics** section, step 10, I had to edit `map_topics_clients.py` and change `#!/usr/bin/env python` to `#!/usr/bin/env python3`.

- [ ] In the **Connect** section, step 2, the cluster name should be `cluster1`. This name change means all Connect screenshots need to be updated as well.

- [ ] In the **KSQL** section, step 2, the ksql application name should be `cluster1`. This name change means all ksql screenshots need to be updated as well.

- [ ] In the **KSQL** section, step 5, add `(If you are using the KSQL CLI, at the ksql> prompt type DESCRIBE WIKIPEDIA;)`.

- [ ] In the **KSQL** section, step 10 is followed by step 1.

- [ ] In the **Consumers** section, steps 3 and 5 mention `_confluent-ksql-default_query_CSAS_WIKIPEDIABOT_0` but the screenshots show `_confluent-ksql-default_query_CSAS_WIKIPEDIANOBOT_0`.

- [ ] In the **Consumers** section, update step 10 to read:

10. From the **Brokers -> Production** view, click on a point in the `Request latency` line graph to view a breakdown of latencies through the entire request lifecycle.

- [ ] In the **Replicator** section, step 5 says "Click on `Consumer Groups` for the topic `wikipedia.parsed` and observe that one of the consumer groups is called `connect-replicator`." I could not figure out how to do this based upon the instruction.

- [ ] In the **Replicator** section, step 8, there is a significant lag between the time `play` is clicked and when C3 shows any reflection of this. It might be worth adding a note indicating to be patient for the change to be reflected. I didn't recognize the delay was present when I first completed this step and clicked `play` and `pause` multiple times. This ended up causing the connector to fail.

docker-compose.yml
```diff
@@ -1,21 +1,49 @@
...
+    image: osixia/openldap:1.3.0
+    hostname: openldap
+    container_name: openldap
     ports:
-      - "2181:2181"
+        - 389:389
+    environment:
+        LDAP_ORGANISATION: "Confluent"
+        LDAP_DOMAIN: "confluent.io"
+        LDAP_BASE_DN: "dc=confluent,dc=io"
```
javabrett: Not suggesting this is wrong, but wondering if it would be better to use a fake org name here, or a modification of Confluent e.g. `confluentdemo`, to save any possible confusion.

Also we are using the container-default admin password - a test/demo system so unlikely to be an issue - but maybe the LDAP instance is accessible remotely and someone accidentally stores a valuable credential?  Trade-off I guess between extra security diligence and convenience.  If generating a password it could be relayed to container stdout.

Accepted that this is a demo, non-production config so take with a grain of salt.

ybyzek: > Not suggesting this is wrong, but wondering if it would be better to use a fake org name here, or a modification of Confluent e.g. confluentdemo, to save any possible confusion.

Great suggestion, changed to `confluentdemo`

PR #166, opened 2020-03-10
Expose plaintext ports for brokers and schema registry

pkleindl:
@rspurgeon This allows connection from the local machine, e.g. I use https://kafka.esque.at/ to access Kafka


pkleindl:
@ybyzek Thanks for the response.
I took a quick look at the PR and it seems the schema registry is not exposed via HTTP
Could you please add that, our tool (and others too) can use that automatically when deserializing.

https://github.com/confluentinc/cp-demo/pull/164/files#diff-4e5e90c6228fd48698d074241c2ba760R627
https://github.com/confluentinc/cp-demo/pull/164/files#diff-4e5e90c6228fd48698d074241c2ba760R621

PR #167, opened 2020-03-11
Changed verify_installed Bash function

dhoard:
Changed verify_installed Bash function to reference python3 executable. On most default installations python is python version 2.


dhoard:
Closing since other script changes should be made to reference python3.

PR #171, opened 2020-03-12
DEVX-1600: Fail fast: if time expires for any of the wait times

ybyzek:


rspurgeon:
If a failure occurs should we cleanup resources instead of just exiting?  I'm not sure as the remaining info could be helpful for debugging, but it's also somewhat poor form to leave stuff half baked on someone's machine.  Mayben we print an error message with the cleanup command so they know how?

ybyzek:
@rspurgeon the reasoning to NOT clean up resources is to enable the user to troubleshoot: get error logs, poke around, etc.

rspurgeon:
> @rspurgeon the reasoning to NOT clean up resources is to enable the user to troubleshoot: get error logs, poke around, etc.

Right, I mentioned that, but we should provide a message IMO so they know

ybyzek:
@rspurgeon After a successful demo run, the script itself also doesn't print how to clean up resources, so it's consistent.  The docs *do* tell users how to cleanup.

rspurgeon:
> @rspurgeon After a successful demo run, the script itself also doesn't print how to clean up resources, so it's consistent. The docs _do_ tell users how to cleanup.

In the case of a successful run the services are running properly and provide instructions on how to continue (click this URL to see C3), here we have a timeout issue but don't provide details in the error message what to do next.

PR #172, opened 2020-03-16
DEVX-1612: Logic broken for retry/attempt

ybyzek:


scripts/helper/functions.sh
```diff
@@ -1,20 +1,21 @@
...
     local -r cmd="$@"
-    local -i attempt_num=1
+
+    local -i sleep_interval=5
+    local -i curr_wait=0
 
     until $cmd
     do
-        if (( attempt_num == max_attempts ))
+        if (( curr_wait > max_wait ))
```
rspurgeon: nit: maybe should be `>=` ?

for example I ask it to wait 5 seconds but it waits 10 with:

```
time retry 5 host_check_mds_up
5
..ERROR: Failed after 10 seconds. Please troubleshoot and run again.
```

ybyzek: Ack and resolved in latest commit https://github.com/confluentinc/cp-demo/pull/172/commits/574656fb81283c8120e50fc9ab83f776929e530f

PR #186, opened 2020-04-06
DEVX-1683: tune configuration for ksqlDB logging

ybyzek:


docker-compose.yml
```diff
@@ -747,7 +747,6 @@ services:
       KSQL_KSQL_INTERNAL_TOPIC_REPLICAS: 2
       KSQL_KSQL_SINK_REPLICAS: 2
 
-      KSQL_LOG4J_OPTS: "-Dlog4j.configuration=file:/tmp/helper/log4j-secure.properties"
```
mikebin, line 750: Are we losing the KSQL Processing log with this change?

ybyzek: @mikebin very good catch!  Sorry about that.  IIUC then no change is needed to cp-demo actually (for DEVX-1683)

ybyzek: I'm going to close this jira, but if I've missed something, please let me know and I will re-open.

PR #195, opened 2020-04-13
DEVX-1589: CP demo Connect worker should not have producer, consumer, or admin principals configured

ybyzek:


docker-compose.yml
```diff
@@ -410,11 +410,6 @@ services:
       # Producer
       CONNECT_PRODUCER_SASL_MECHANISM: 'OAUTHBEARER'
```
mikebin, line 411: Should these properties (`CONNECT_PRODUCER_SASL_MECHANISM` and `CONNECT_PRODUCER_SASL_LOGIN_CALLBACK_HANDLER_CLASS`) also be removed from the worker, and moved to the connectors to avoid separating related parts of the security config between worker and connector?

mikebin: Same goes for the Consumer and Admin configs below.

ybyzek: @mikebin technically it can go either way, so the question is what logically makes sense.

Pro for moving to connector:
- Keep similar configs together

Con for moving to connector:
- Duplication

Given that it's an RBAC environment, it's not possible (or correct me as needed) for connectors to NOT use `TokenUserLoginCallbackHandler`, so I might lean towards keeping it on the worker to avoid duplication -- i.e., this isn't similar to Avro/SR configs where some connectors might use it and others might not so it makes sense to keep those all together.  But I understand that it feels weird splitting it up, so I'm actually on the fence.  @C0urante WDYT?

mikebin: Actually, I'm not sure token auth is required in the case of Connect. Seems like normal client authentication mechanisms could be used in this case, since identity propagation/impersonation is not required. @C0urante I'm sure can provide a more definitive answer.

C0urante: @mikebin is correct; all auth mechanisms are supported. However, I like the defaults here since they play nicely with the `principal.service.name` and `principal.service.password` configs that we've set up to help users provide credentials for their connectors. Those two configs are taken and translated into `consumer.override.sasl.jaas.config`, `producer.override.sasl.jaas.config`, and `admin.override.sasl.jaas.config` properties under the hood, but `consumer.override.security.protocol`, `consumer.override.sasl.mechanism`, `consumer.override.sasl.login.callback.handler.class`, etc. are _not_ configured. So I think it'd be a better OOTB UX to leave these in here for now and only remove the `consumer.sasl.jaas.config`, `producer.sasl.jaas.config`, and `admin.sasl.jaas.config` properties from the worker file.

ybyzek:
@C0urante yep, tested and verified.

PR #207, opened 2020-04-23
DEVX-1741: Increase wait time for subject wikipedia.parsed-value to b…

ybyzek:
…e registered

1. Increase wait time for subject to be registered in Schema Registry
2. Move up to introduce small time delay with hope to cover ES not ready (separate issue observed once)


daveshook:
LGTM

PR #208, opened 2020-04-23
DEVX-1769: Enable certs-create-per-user.sh to be run from any location

ybyzek:


scripts/security/certs-create-per-user.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+ROOT_CA_DIR=$(realpath $(dirname "${BASH_SOURCE[0]}"))
```
ybyzek, line 3: @jeanlouisboudart IIUC the goal of this is to provide the real path to this directory, e.g.

```
/Users/yeva/git/cp-demo/scripts/security
```

There is another place in this demo with an example of a command that does this as well, https://github.com/confluentinc/cp-demo/blob/5.5.0-post/scripts/start.sh#L3

```
```

For consistency, we should use the same command in both places.  Do you have concern if we changed this to:

```
ROOT_CA_DIR ="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```

jeanlouisboudart: That's exactly the goal.
I'll make that change tomorrow.

scripts/security/certs-create-per-user.sh
```diff
@@ -1,5 +1,7 @@
...
+echo $ROOT_CA_DIR
```
ybyzek, line 4: This looks a bit messy in the output.  Options:

1. Delete this line
2. Add better echo, e.g. `echo "Writing credential files in path $ROOT_CA_DIR"`

ybyzek:
Need to update https://github.com/confluentinc/jmx-monitoring-stacks/blob/5.5.0-post/README.md with new instructions

scripts/security/certs-create-per-user.sh
jeanlouisboudart: I don't think we can use this trick here as it change the current directory to that particular folder :).

As such later in the start.sh everything is broken :)

scripts/security/certs-create-per-user.sh
jeanlouisboudart: Done in latest commit, i've remove this unnecessary log

rspurgeon:
What is meant by?

>  Introduce a DOCKER_COMPOSE_OPTS variable to allow to have cp-demo extensions in another repository

rspurgeon:
Please see this comment regarding the need for a new DOCKER_COMPOSE_OPTS value

https://github.com/confluentinc/jmx-monitoring-stacks/pull/2#discussion_r414651443

jeanlouisboudart:
I've reverted the change @rspurgeon, thanks for the pointer @rspurgeon I wasn't aware of this COMPOSE_FILE variable.

scripts/security/certs-create-per-user.sh
jeanlouisboudart: is this Ok for you @ybyzek ?

ybyzek: @jeanlouisboudart the only outstanding concern of `realpath` (aside from lack of consistency) is that it's non-standard.  Some people have `dirname` but do not have `realpath` (I checked with another user), so it results in a new added dependency to run this demo (e.g. `brew install coreutils` on macos) -- in contrast, I *suspect* (but don't know for sure) that `dirname` is more standard.  Can we try to make `dirname` do what you want it to do?

cc: @rspurgeon for additional ideas.

jeanlouisboudart: I understand your concerns now @ybyzek and thanks for pointing this.
I removed the dependency and added a helper script doing exactly what we need.

scripts/helper/fullpath.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+fullpath ()
+{
+  fullPath=$(cd $1 && pwd -P)
+  echo $fullPath
+  cd $OLDPWD
```
rspurgeon, line 7: I do not believe the final `cd` is needed here and maybe has an undesired side effect?   Since line 5 is ran in a subshell the current directory is not changed

PR #209, opened 2020-04-24
DEVX-1717: Update C3 screenshots in cp-demo in 5.5.0 with new C3 branding

daveshook:
This is includes all but one of the required CP 5.5 screenshots. The Consumption Difference alert was never triggered.

It also includes rewording of a few steps.


daveshook:
I was able to create the last required screenshot.

daveshook:
I ran CP Demo again and the Consumption Alert was triggered so I captured the screenshot and replaced the one that I "created".

PR #211, opened 2020-05-01
Updated the training-tools container image to 5.5

javabrett:
Note: this rev looks like it could be easily-missed since versions are externalised to `config.env` e.g. `CONFLUENT_DOCKER_TAG=5.5.0`.  I wondered if another variable could be added close-by, or even co-opt `CONFLUENT_SHORT=5.5` so that they would be updated together - thoughts?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,7 @@ services:
       - ./scripts/security:/etc/kafka/secrets
 
   tools:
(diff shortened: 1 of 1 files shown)


ybyzek:
@javabrett at this time, we cannot tie them together because a different team creates/manages/publishes this Docker image, and as happened with this latest release cut, it can happen after GA.  So it has to be manually updated, for the time being.

PR #212, opened 2020-05-01
DEVX-1719 adds documentation to docker-compose files

rspurgeon:
explaining our usage of environment variables here.


rspurgeon:
@ybyzek I will address the copy edits from examples PR here as well

PR #220, opened 2020-06-03
DEVX-1843: cp-demo repo: replace KSQL with ksqlDB

ybyzek:


scripts/helper/create-role-bindings.sh
```diff
@@ -22,9 +22,9 @@ CONNECT_ADMIN="User:connectAdmin"
 CONNECTOR_SUBMITTER="User:connectorSubmitter"
 CONNECTOR_PRINCIPAL="User:connectorSA"
 SR_PRINCIPAL="User:schemaregistryUser"
-KSQL_ADMIN="User:ksqlDBAdmin"
-KSQL_USER="User:ksqlDBUser"
-KSQL_SERVER="User:ksqldbserver"
+KSQLDB_ADMIN="User:ksqlDBAdmin"
+KSQLDB_USER="User:ksqlDBUser"
+KSQLDB_SERVER="User:ksqldbserver"
```
mikebin, line 27: Not really related to this change, but is there a reason this user has lowercase `db` in the username, while the other users have uppercase `DB`?

ybyzek: Good catch.  Fixing now.

PR #225, opened 2020-06-29
[DEVX-1887] updated cp-demo to use license manager instead of ZK for trial license - SR & Rest

sdanduConf:
In order for SR and Rest Proxy to use the confluent license manager for trial license management, they need read/write access on the _confluent_license topic. This adds those role bindings.

Testing: Confirmed before this change that rest / SR failed to start due to lack of topic access and with this change both SR and Rest start fine (get past licensing) - Demo Not fully tested e-2-e

One additional change added here to modify ksql authorization to support a recent switch to vert.x auth

Testing: Internal error before this change due jetty no longer being available. Authorization was successful after this change.

### DIFF
--- a/scripts/helper/create-role-bindings.sh
+++ b/scripts/helper/create-role-bindings.sh
@@ -14,6 +14,7 @@ CONNECT=connect-cluster
 SR=schema-registry
 KSQLDB=ksql-cluster
 C3=c3-cluster
+LICENSE_RESOURCE="Topic:_confluent-license"
 
 SUPER_USER=superUser
 SUPER_USER_PASSWORD=superUser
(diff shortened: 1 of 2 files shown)


scripts/helper/create-role-bindings.sh
```diff
@@ -77,6 +78,24 @@ do
         --kafka-cluster-id $KAFKA_CLUSTER_ID
 done
 
+for resource in Topic:_confluent-license
```
ybyzek, line 81: @sdanduConf with only one resource, can we remove the for loop?

e.g.

```
 confluent iam rolebinding create \
         --principal $SR_PRINCIPAL \
         --role DeveloperRead \
         --resource Topic:_confluent-license \
         --kafka-cluster-id $KAFKA_CLUSTER_ID
```

or

make the for loop to cover both roles.

Comment here applies to both SR and REST Proxy

sdanduConf: @ybyzek updated

PR #227, opened 2020-06-29
DEVX-1893: Change elasticsearch and other non-Replicator connectors t…

ybyzek:
…o use latest


ybyzek:
@sdanduConf verified with a run of cp-demo

PR #236, opened 2020-07-28
DEVX-1957: add promo code to docs

ybyzek:
@rspurgeon docs change only

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -71,7 +71,7 @@ Docker
 This is a Docker environment and has all services running on one host.
 It is meant exclusively to easily demo the |CP|, but do not deploy all |cp| services on a single host in production.
 
-Also, in production, |c3| should be deployed with a valid license and with its own dedicated metrics cluster, separate from the cluster with production traffic.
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -61,26 +65,48 @@ Demo validated with:
...
 -  jq
 
-.. note:: If you prefer other non-Docker demos, please go to `confluentinc/examples GitHub repository <https://github.com/confluentinc/examples>`__.
+Docker
+------
+
+This is a Docker environment and has all services running on one host.
+It is meant exclusively to easily demo the |CP|, but do not deploy all |cp| services on a single host in production.
+
+Also, in production, |c3| should be deployed with a valid license and with its own dedicated metrics cluster, separate from the cluster with production traffic.
```
rspurgeon, line 74: Maybe lean towards "data" or "events" vs "traffic" which implies something less about Kafka IMO
```suggestion
Also, in production, |c3| should be deployed with a valid license and with its own dedicated metrics cluster, separate from the cluster with production data.
```

PR #239, opened 2020-08-19
Enable EmbeddedKafkaRest with RBAC on kafka1

rnpridgeon:


rnpridgeon:
Note: Relies on `REPOSITORY=368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc` being set in `env_files/config.env`

rnpridgeon:
To verify: 

 curl --user alice:alice-secret http://localhost:8091/kafka/v3/clusters

{"kind":"KafkaClusterList","metadata":{"self":"http://localhost:8091/kafka/v3/clusters","next":null},"data":[{"kind":"KafkaCluster","metadata":{"self":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA","resource_name":"crn:///kafka=OUM6maRLTH2dPEM7MQZIlA"},"cluster_id":"OUM6maRLTH2dPEM7MQZIlA","controller":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/brokers/2"},"acls":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/acls"},"brokers":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/brokers"},"broker_configs":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/broker-configs"},"consumer_groups":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/consumer-groups"},"topics":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/topics"},"partition_reassignments":{"related":"http://localhost:8091/kafka/v3/clusters/OUM6maRLTH2dPEM7MQZIlA/topics/-/partitions/-/reassignment"}}]}


 curl --user alice:alice-bad-secret http://localhost:8091/kafka/v3/clusters

<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 401 Unauthorized</title>
</head>
<body><h2>HTTP ERROR 401 Unauthorized</h2>
<table>
<tr><th>URI:</th><td>/kafka/v3/clusters</td></tr>
<tr><th>STATUS:</th><td>401</td></tr>
<tr><th>MESSAGE:</th><td>Unauthorized</td></tr>
<tr><th>SERVLET:</th><td>default</td></tr>
</table>

</body>
</html>

docker-compose.yml
```diff
@@ -204,6 +204,37 @@ services:
...
+      KAFKA_KAFKA_REST_CLIENT_SSL_TRUSTSTORE_PASSWORD: confluent
+      KAFKA_KAFKA_REST_CLIENT_SSL_KEYSTORE_LOCATION: /etc/kafka/secrets/kafka.kafka1.keystore.jks
+      KAFKA_KAFKA_REST_CLIENT_SSL_KEYSTORE_PASSWORD: confluent
+      KAFKA_KAFKA_REST_CLIENT_SSL_KEY_PASSWORD: confluent
+      # EmbeddedKafkaRest: HTTP Auth Configuration
+      KAFKA_KAFKA_REST_KAFKA_REST_RESOURCE_EXTENSION_CLASS: io.confluent.kafkarest.security.KafkaRestSecurityResourceExtension
+      KAFKA_KAFKA_REST_REST_SERVLET_INITIALIZOR_CLASSES: io.confluent.common.security.jetty.initializer.InstallBearerOrBasicSecurityHandler
+      KAFKA_KAFKA_REST_PUBLIC_KEY_PATH: /tmp/conf/public.pem
+      # EmbeddedKafkaRest: MDS (HTTP) Client configuration
+      KAFKA_KAFKA_REST_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS: 60000
```
ybyzek, line 223: @rnpridgeon why is this required: `KAFKA_KAFKA_REST_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS` and `KAFKA_KAFKA_REST_CLIENT_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS`?

rnpridgeon: Really neither are necessarily required, the defaults are usually fine, but you set them in the stand alone RP config so I preserved them here. 

As for why there are two:
KAFKA_KAFKA_REST_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS is for KafkaRest itself. 
KAFKA_KAFKA_REST_CLIENT_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS is for the base client config use when creating clients in response to user requests.

rnpridgeon:
Note about docs inclusion:
1. Currently references the v3 API docs which are in staging; this will need to be updated when they exist 
2. I have no idea how to render the rst file to check for correctness

ybyzek:
@rnpridgeon I'll review this PR now, thanks.  

Meanwhile

1. ack and thank you.
2. Please see https://github.com/confluentinc/docs/pull/5517 for staging docs

docs/index.rst
```diff
@@ -697,7 +697,7 @@ The security in place between |sr| and the end clients, e.g. ``appSA``, is as fo
    .. sourcecode:: bash
 
       # First get the KAFKA_CLUSTER_ID
-      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+     KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v3/clusters | jq ".data[0].cluster_id"")
```
ybyzek, line 700: @rnpridgeon cp-demo code continues to refer to the `v1/metadata/id` endpoint instead of `v3/clusters` endpoint.

1. For consistency, it should all be the same.  Need to decide on one or the other.
2. Which one should it be in 6.0.x?  What's the difference between the two endpoints, when should a user hit one versus the other?

docs/index.rst
```diff
@@ -793,6 +793,9 @@ Confluent REST Proxy
 --------------------
 
 The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/index.html>`__  is running for optional client access.
+It may run in standalone mode or in embedded most hosted on the broker.
+While in embedded mode only the .. _V3 APIs: http://staging-docs.confluent.io.s3-website-us-west-2.amazonaws.com/current/kafka-rest/api.html#crest-api-v3 are supported this time.
```
ybyzek, line 797:
```suggestion
While in embedded mode only the :ref:`rest-proxy-v3` are supported this time.
```

docs/index.rst
```diff
@@ -793,6 +793,9 @@ Confluent REST Proxy
 --------------------
 
 The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/index.html>`__  is running for optional client access.
+It may run in standalone mode or in embedded most hosted on the broker.
```
ybyzek, line 796:
```suggestion
This demo showcases |crest| in two modes: (1) running as a standalone service and (2) embedded on the |ak| brokers.
```

docs/index.rst
```diff
@@ -793,6 +793,9 @@ Confluent REST Proxy
...
+While in embedded mode only the .. _V3 APIs: http://staging-docs.confluent.io.s3-website-us-west-2.amazonaws.com/current/kafka-rest/api.html#crest-api-v3 are supported this time.
+Additionally each API's URL must include the prefix ``/kafka``
```
ybyzek, line 798: Delete this line

docs/index.rst
```diff
@@ -924,6 +927,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
 
        docker-compose exec restproxy curl -X DELETE -H "Content-Type: application/vnd.kafka.v2+json" --cert /etc/kafka/secrets/restproxy.certificate.pem --key /etc/kafka/secrets/restproxy.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt -u appSA:appSA https://restproxy:8086/consumers/my_avro_consumer/instances/my_consumer_instance
 
+#. Create a role binding for the client providing it total ownership of topic ``dev_users``.
```
ybyzek, line 930:
```suggestion
#. Now you will use the |crest| that is embedded on the |ak| brokers.  Create a role binding for the client to be granted ``ResourceOwner`` role for the topic ``dev_users``.
```

docs/index.rst
```diff
@@ -924,6 +927,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v3/clusters | jq ".data[0].cluster_id"")
+
+      # Then create the role binding for the topic ``dev_users``
+      docker-compose exec tools bash -c "confluent iam rolebinding create \
+          --principal User:appSA \
+          --role ResourceOwner \
+          --resource Topic:usersq \
+          --kafka-cluster-id $KAFKA_CLUSTER_ID"
+
+#. Create the topic ``dev_users``.
```
ybyzek, line 944:
```suggestion
#. Create the topic ``dev_users`` with the embedded |crest|.
```

docker-compose.yml
```diff
@@ -205,6 +205,29 @@ services:
       KAFKA_LDAP_USER_NAME_ATTRIBUTE: uid
       KAFKA_LDAP_USER_OBJECT_CLASS: inetOrgPerson
 
+      # ConfluentTelemetryReporter: Disable
+      KAFKA_CONFLUENT_TELEMETRY_REPORTER_ENABLE: 'false'
+
+      # EmbeddedKafkaRest: Kafka Client Configuration
```
ybyzek, line 211: To be explicit, can you please set the configuration parameter for the port for the embedded RP? On this and the broker below

docs/index.rst
```diff
@@ -793,6 +793,9 @@ Confluent REST Proxy
 --------------------
 
 The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/index.html>`__  is running for optional client access.
+It may run in standalone mode or in embedded most hosted on the broker.
+While in embedded mode only the .. _V3 APIs: http://staging-docs.confluent.io.s3-website-us-west-2.amazonaws.com/current/kafka-rest/api.html#crest-api-v3 are supported this time.
+Additionally each API's URL must include the prefix ``/kafka``
 
 #. Use the |crest|, which is listening for HTTPS on port 8086, to try to produce a message to the topic ``users``, referencing schema id ``7``. This schema was registered in |sr| in the previous section. It should fail due to an authorization error.
```
ybyzek, line 800:
```suggestion
#. Use the standalone |crest|, which is listening for HTTPS on port 8086, to try to produce a message to the topic ``users``, referencing schema id ``7``. This schema was registered in |sr| in the previous section. It should fail due to an authorization error.
```

docs/index.rst
rnpridgeon: ERP is on by default, MDS isn't. So I'd personally use the new one. I thought I got them all but alas I guess I missed some.

docker-compose.yml
```diff
@@ -205,6 +205,29 @@ services:
       KAFKA_LDAP_USER_NAME_ATTRIBUTE: uid
       KAFKA_LDAP_USER_OBJECT_CLASS: inetOrgPerson
 
+      # ConfluentTelemetryReporter: Disable
+      KAFKA_CONFLUENT_TELEMETRY_REPORTER_ENABLE: 'false'
```
ybyzek, line 209: @rnpridgeon 

1. Why is telemetry reporter disabled?
2. Should it be enabled in cp-demo?  (cc: @sdanduConf )

docs/index.rst
ybyzek: 1. `http://localhost:8091/v1/metadata/id` is also used in the scripts themselves, so those would need to change as well.
2. What are the other consequences of changing the endpoint?  What if the embedded RP dies?

docker-compose.yml
rnpridgeon: This was a work around because the internal exporter topic had a default replication factor of 3 and attempts to override it did not work. I think this has been fixed and it can be removed

docker-compose.yml
rnpridgeon: We can either set `KAFKA_CONFLUENT_METADATA_SERVER_LISTENERS` or `KAFKA_CONFLUENT_HTTP_SERVER` but not both. This is a consequence of MDS and KafkaRest sharing the same underlying HTTP server. The former only still exists for backwards compatibility reasons. I'm fine with fixing forward but this would make MDS less "explicit" so it's up to you .

docs/index.rst
rnpridgeon: 1. okay I'm impartial but if we want to demonstrate the new API we should use the new one
2. It goes away; this is true for the former UI if MDS dies though so using the former implementation isn't more robust.

ybyzek:
@rnpridgeon I know we chatted about this in Slack, but just wanted to post here that the latest image does not have the KafkaHttpServer DOA symptoms ;)

docker-compose.yml
ybyzek: @rnpridgeon I'm confused then because the example in the docs that you added operates on endpoint `http://localhost:8082/v3/clusters/${KAFKA_CLUSTER_ID}/topics`, and 8082 appears to be the standalone RP, not the ERP?  The goal is to differentiate what action is being taken on which RP.

docs/index.rst
ybyzek: Ok, in that case, please update all instances (see scripts):

```
scripts/validate/validate_rest_proxy.sh:KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
scripts/validate/validate_bindings.sh:KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
scripts/helper/functions.sh:  KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```

docker-compose.yml
ybyzek: @rnpridgeon cool, so will you be flipping this to `true`?

docker-compose.yml
ybyzek: Nevermind, standalone RP uses 8086, and the embedded RP uses 8082.  The confusion was resulting from the docker-compose output which shows 8082.  As a remediation, at the top of this RP section, it should state which port corresponds to which instance.

rnpridgeon: Standalone RP doesn't share an underlying HTTP server instance with MDS; ERP does. It's not really a docker compose thing or a RestProxy thing it's an implementation detail with regard to how the RestProxy servlet is hosted. That said sure we can add a note at the top of the section

docker-compose.yml
rnpridgeon: I'll just remove it entirely

docker-compose.yml
```diff
@@ -343,6 +363,28 @@ services:
...
```
ybyzek, line 367: @rnpridgeon Looks like this one was missed for deletion?  (see https://github.com/confluentinc/cp-demo/pull/239/commits/675ae669d75814c9fe4e829c268ec77f2d8c0486 for deletion on other broker config)

docker-compose.yml
```diff
@@ -343,6 +363,28 @@ services:
...
+      KAFKA_KAFKA_REST_CLIENT_SSL_KEYSTORE_PASSWORD: confluent
+      KAFKA_KAFKA_REST_CLIENT_SSL_KEY_PASSWORD: confluent
+      # EmbeddedKafkaRest: HTTP Auth Configuration
+      KAFKA_KAFKA_REST_KAFKA_REST_RESOURCE_EXTENSION_CLASS: io.confluent.kafkarest.security.KafkaRestSecurityResourceExtension
+      KAFKA_KAFKA_REST_REST_SERVLET_INITIALIZOR_CLASSES: io.confluent.common.security.jetty.initializer.InstallBearerOrBasicSecurityHandler
+      KAFKA_KAFKA_REST_PUBLIC_KEY_PATH: /tmp/conf/public.pem
+      # EmbeddedKafkaRest: MDS (HTTP) Client configuration
+      KAFKA_KAFKA_REST_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS: 60000
+      KAFKA_KAFKA_REST_CLIENT_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS: 60000
+      KAFKA_KAFKA_REST_CONFLUENT_METADATA_BOOTSTRAP_SERVER_URLS: http://kafka1:8091,http://kafka2:8092
```
ybyzek, line 384: Just to confirm, is this the right config, whereby the ERP in broker 2 may be configured with bootstrap server for broker 1?

docs/index.rst
```diff
@@ -793,8 +793,10 @@ Confluent REST Proxy
 --------------------
 
 The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/index.html>`__  is running for optional client access.
+This demo showcases |crest| in two modes: (1) running as a standalone service and (2) embedded on the |ak| brokers.
+While in embedded mode |crest| listens for requests on port 8092 providing only the :ref:`rest-proxy-v3` are supported this time.
```
ybyzek, line 797: There are two ERP instances, one on broker1 and one on broker2.  In Docker they shouldn't both be listening on port 8092.  Can it be differentiated?

docker-compose.yml
rnpridgeon: Yes, you want the HTTP client to have multiple MDS URLs in case one is down it can bootstrap to the other. It works a lot like Kafka client bootstrap configs.

docs/index.rst
rnpridgeon: MDS and KafkaRest share the same HTTP server so they have the same listeners. I can add the full address for both

docs/index.rst
```diff
@@ -924,6 +926,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+          --kafka-cluster-id $KAFKA_CLUSTER_ID"
+
+#. Create the topic ``dev_users`` with the embedded |crest|.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v3/clusters | jq ".data[0].cluster_id")
+
+      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" --cert /etc/kafka/secrets/restproxy.certificate.pem --key /etc/kafka/secrets/restproxy.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt -u appSA:appS "http://localhost:8082/v3/clusters/${KAFKA_CLUSTER_ID}/topics" -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}"
```
ybyzek, line 950: This is hitting port 8082, isn't that contradictory with the earlier statement:

```
While in embedded mode |crest| listens for requests on http://kafka1:8091 and http://kafka2:8092 providing only the :ref:`rest-proxy-v3` are supported this time.
```

rnpridgeon: Right you are. Does it just make sense to use the v3 APIs on the standalone leaving it up to the reader to try v3 apis against the embedded instance at their leisure.

ybyzek: IIUC v3 doesn't support produce/consume, so the standalone can't use v3?  If that is correct, then I would suggest:

a) standalone uses v1 with produce/consume -- requires no change to existing docs
b) ERP uses v3 with just topic creation

rigelbm: a) standalone uses V2 with produce/consume (V1 is gone in 6.0)

ybyzek: @rigelbm correction, cp-demo example is already using v2 for the standalone (not v1).  If 6.0 supports v2 and v3, then this PR can optionally update all the commands to use v3 syntax for standalone.

Still open issue is reconciling the ports: why is the ERP example using 8082 when the added writing says 8091 and 8092

rnpridgeon: It's a typo again my suggestion is to not mix the two in the examples but instead mention that ERP listens on 809[1-2] in the intro if you want to execute v3 requests against it.

ybyzek: My strong leaning is that if this PR adds ERP to the configuration, then the PR should also add an exercise to execute statements to the ERP.

docs/index.rst
```diff
@@ -793,8 +793,10 @@ Confluent REST Proxy
 --------------------
 
 The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/index.html>`__  is running for optional client access.
+This demo showcases |crest| in two modes: (1) running as a standalone service and (2) embedded on the |ak| brokers.
```
ybyzek, line 796:
```suggestion
This demo showcases |crest| in two modes: (1) running as a standalone service listening for HTTPS requests on port 8086,  and (2) embedded on the |ak| brokers listening for HTTP requests on port 8091 on ``kafka1`` and on port 8092 on ``kafka2`` (the port is shared with the MDS listener)
```

docs/index.rst
```diff
@@ -793,8 +793,10 @@ Confluent REST Proxy
...
+While in embedded mode |crest| listens for requests on http://kafka1:8091/kafka and http://kafka2:8092/kafka providing only the :ref:`rest-proxy-v3` are supported this time.
```
ybyzek, line 797: Delete this line

docs/index.rst
```diff
@@ -924,6 +926,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
 
        docker-compose exec restproxy curl -X DELETE -H "Content-Type: application/vnd.kafka.v2+json" --cert /etc/kafka/secrets/restproxy.certificate.pem --key /etc/kafka/secrets/restproxy.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt -u appSA:appSA https://restproxy:8086/consumers/my_avro_consumer/instances/my_consumer_instance
 
+#. Now you will use the |crest| that is embedded on the |ak| brokers.  Create a role binding for the client to be granted ``ResourceOwner`` role for the topic ``dev_users``.
```
ybyzek, line 929:
```suggestion
#. For the next few steps, use the |crest| that is embedded on the |ak| brokers. Only :ref:`rest-proxy-v3` is supported this time.  Create a role binding for the client to be granted ``ResourceOwner`` role for the topic ``dev_users``.
```

docs/index.rst
```diff
@@ -793,8 +793,10 @@ Confluent REST Proxy
 --------------------
 
 The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/index.html>`__  is running for optional client access.
+This demo showcases |crest| in two modes: (1) running as a standalone service and (2) embedded on the |ak| brokers.
+While in embedded mode |crest| listens for requests on http://kafka1:8091/kafka and http://kafka2:8092/kafka providing only the :ref:`rest-proxy-v3` are supported this time.
 
-#. Use the |crest|, which is listening for HTTPS on port 8086, to try to produce a message to the topic ``users``, referencing schema id ``7``. This schema was registered in |sr| in the previous section. It should fail due to an authorization error.
+#. Use the standalone |crest|, which is listening for HTTPS on port 8086, to try to produce a message to the topic ``users``, referencing schema id ``7``. This schema was registered in |sr| in the previous section. It should fail due to an authorization error.
```
ybyzek, line 799:
```suggestion
#. Use the standalone |crest| to try to produce a message to the topic ``users``, referencing schema id ``7``. This schema was registered in |sr| in the previous section. It should fail due to an authorization error.
```

scripts/helper/functions.sh
```diff
@@ -52,7 +52,7 @@ preflight_checks()
 
 get_kafka_cluster_id_from_container()
 {
-  KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
+  KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/kafka/v3/clusters | jq ".data[0].cluster_id")
```
ybyzek, line 55: This breaks the demo with the following failure:

```
[8:40:34] ~/git/cp-demo(PR-239) ✗: curl -s http://localhost:8091/kafka/v3/clusters 
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 401 Unauthorized</title>
</head>
<body><h2>HTTP ERROR 401 Unauthorized</h2>
<table>
<tr><th>URI:</th><td>/kafka/v3/clusters</td></tr>
<tr><th>STATUS:</th><td>401</td></tr>
<tr><th>MESSAGE:</th><td>Unauthorized</td></tr>
<tr><th>SERVLET:</th><td>default</td></tr>
</table>

</body>
</html>
```

The original works:

```
[8:40:32] ~/git/cp-demo(PR-239) ✗: curl -s http://localhost:8091/v1/metadata/id    
{"id":"QTtyaXufRuOfIUCj2JIcLw","scope":{"path":[],"clusters":{"kafka-cluster":"QTtyaXufRuOfIUCj2JIcLw"}}}  
```

For all instances (code & docs) where the Kafka cluster ID is determined, can we restore it to the original?

rnpridgeon: hm that is odd because it certainly worked before. I'll back it out for now and investigate separately. May be a regression that needs to be addressed in 6.0.1.

docs/index.rst
rnpridgeon: Both examples point to kafka1:8091 now

docs/index.rst
```diff
@@ -811,7 +812,7 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
    .. sourcecode:: bash
 
       # First get the KAFKA_CLUSTER_ID
-      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+      KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```
ybyzek, line 815:
```suggestion
      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

docs/index.rst
```diff
@@ -867,7 +868,7 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
```
ybyzek, line 871:
```suggestion
      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

docs/index.rst
```diff
@@ -895,7 +896,7 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
```
ybyzek, line 899:
```suggestion
      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
 
        docker-compose exec restproxy curl -X DELETE -H "Content-Type: application/vnd.kafka.v2+json" --cert /etc/kafka/secrets/restproxy.certificate.pem --key /etc/kafka/secrets/restproxy.key --tlsv1.2 --cacert /etc/kafka/secrets/snakeoil-ca-1.crt -u appSA:appSA https://restproxy:8086/consumers/my_avro_consumer/instances/my_consumer_instance
 
+#. For the next few steps, use the |crest| that is embedded on the |ak| brokers. Only :ref:`rest-proxy-v3` is supported this time.  Create a role binding for the client to be granted ``ResourceOwner`` role for the topic ``dev_users``.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```
ybyzek, line 933:
```suggestion
      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+          --role ResourceOwner \
+          --resource Topic:usersq \
+          --kafka-cluster-id $KAFKA_CLUSTER_ID"
+
+#. Create the topic ``dev_users`` with embedded |crest|.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```
ybyzek, line 947:
```suggestion
      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

scripts/validate/validate_rest_proxy.sh
```diff
@@ -2,8 +2,8 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 
-################################## GET KAFKA CLUSTER ID ########################
-KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+################################## GET K AFKA CLUSTER ID ########################
```
ybyzek, line 5:
```suggestion
################################## GET KAFKA CLUSTER ID ########################
```

scripts/validate/validate_bindings.sh
```diff
@@ -5,7 +5,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 source ../helper/functions.sh
 
 ################################## GET KAFKA CLUSTER ID ########################
-KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```
ybyzek, line 8:
```suggestion
KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

scripts/validate/validate_rest_proxy.sh
```diff
@@ -2,8 +2,8 @@
...
+KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```
ybyzek, line 6:
```suggestion
KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

From the host machine, this should be `localhost`, not `kafka1`

ybyzek (changes requested):

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+
+      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appS "https://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}"
+
+#. List topics with embedded |crest| to find the newly created ``dev_users``.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://kafka1:8091/v1/metadata/id | jq -r ".id")
```
rnpridgeon, line 956:
```suggestion
      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
```

docs/index.rst
```diff
@@ -578,7 +578,7 @@ End clients (non-CP clients):
    .. sourcecode:: bash
 
       # First get the KAFKA_CLUSTER_ID
-      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+     KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v3/clusters | jq ".data[0].cluster_id"")
```
ybyzek, line 581: IIUC this won't work due to KREST-19?

docs/index.rst
ybyzek, line 700: IIUC this won't work due to KREST-19?

ybyzek (changes requested):

docs/index.rst
ybyzek: @rnpridgeon can you please revert so we can merge?

docs/index.rst
ybyzek: @rnpridgeon can you please revert so we can merge?

rnpridgeon: seems my local repo and the remote repo diverged causing some funkyness. Did a hard reset on the remote repo and reverted these again. Seems the commit suggested change button maybe more of an annoyance than just commiting the suggestion manually.

docs/index.rst
rnpridgeon: alternatively you could use the kafka rest credentials, for now its reverted.

ybyzek (approved):
@rnpridgeon THANK YOU!  LGTM

ybyzek (changes requested):
Need to address test failures -- See Slack

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+
+      # Then create the role binding for the topic ``dev_users``
+      docker-compose exec tools bash -c "confluent iam rolebinding create \
+          --principal User:appSA \
+          --role ResourceOwner \
+          --resource Topic:usersq \
```
ybyzek, line 939:
```suggestion
          --resource Topic:dev_users \
```

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+          --kafka-cluster-id $KAFKA_CLUSTER_ID"
+
+#. Create the topic ``dev_users`` with embedded |crest|.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+
+      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appS "https://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}"
```
ybyzek, line 949: Results in error:

```
curl: (35) error:1408F10B:SSL routines:ssl3_get_record:wrong version number
```

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+
+#. List topics with embedded |crest| to find the newly created ``dev_users``.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+
+      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appS http://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics
```
ybyzek, line 958: Results in error:

```
{
"servlet":"default",
"message":"Unauthorized",
"url":"/kafka/v3/clusters/$KAFKA_CLUSTER_ID/topics",
"status":"401"
}          
```

Note: `appS` --> `appSA`

ybyzek:
```
[10:21:49] ~/git/cp-demo(PR-239) ✗: openssl s_client -connect localhost:8091
CONNECTED(00000005)
4547165632:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:ssl/record/ssl3_record.c:332:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 5 bytes and written 293 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---

```

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appSA "http://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}"
+
+#. List topics with embedded |crest| to find the newly created ``dev_users``.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+
+      docker-compose exec restproxy curl -X GET -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appSA http://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics
```
ybyzek, line 958:
```suggestion
      docker-compose exec restproxy curl -X GET -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appSA http://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics |jq '.data[].topic_name'
```

docs/index.rst
```diff
@@ -924,6 +925,37 @@ The `Confluent REST Proxy <https://docs.confluent.io/current/kafka-rest/docs/ind
...
+          --kafka-cluster-id $KAFKA_CLUSTER_ID"
+
+#. Create the topic ``dev_users`` with embedded |crest|.
+
+   .. sourcecode:: bash
+
+      # First get the KAFKA_CLUSTER_ID
+      KAFKA_CLUSTER_ID=$(curl -s http://localhost:8091/v1/metadata/id | jq -r ".id")
+
+      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appSA "http://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}"
```
ybyzek, line 949:
```suggestion
      docker-compose exec restproxy curl -X POST -H "Content-Type: application/json" -H "accept: application/json" -u appSA:appSA "http://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}" | jq
```

ybyzek (approved):
LGTM

PR #248, opened 2020-09-23
DEVX-1809: CLI commands: replace references of `--zookeeper` with `--bootstrap-server`

ybyzek:


scripts/start.sh
```diff
@@ -48,6 +48,24 @@ sleep 5
...
+# Workaround for setting min ISR on topic _confluent-metadata-auth
+docker-compose exec kafka1 kafka-configs \
+   --bootstrap-server kafka1:12091 \
+   --entity-type topics \
+   --entity-name _confluent-metadata-auth \
+   --alter \
+   --add-config min.insync.replicas=1
+
+# Create Kafka topic users, using appSA principal
+docker-compose exec kafka1 bash -c 'export KAFKA_LOG4J_OPTS="-Dlog4j.rootLogger=DEBUG,stdout -Dlog4j.logger.kafka=DEBUG,stdout" && kafka-topics \
```
mikebin, line 60: Are these log4j options needed?

scripts/start.sh
```diff
@@ -97,6 +115,30 @@ docker-compose exec connect timeout 3 nc -zv irc.wikimedia.org 6667 || {
   exit 1
 }
 
+# Create Kafka topics with prefix wikipedia, using connectorSA principal
+docker-compose exec kafka1 bash -c 'export KAFKA_LOG4J_OPTS="-Dlog4j.rootLogger=DEBUG,stdout -Dlog4j.logger.kafka=DEBUG,stdout" && kafka-topics \
```
mikebin, line 119: Are these log4j options needed?

scripts/start.sh
```diff
@@ -120,6 +162,18 @@ ${DIR}/dashboard/configure_kibana_dashboard.sh
 echo
 echo
 
+# Create Kafka topics with prefix WIKIPEDIA or EN_WIKIPEDIA, using ksqlDBUser principal
+for t in WIKIPEDIABOT WIKIPEDIANOBOT EN_WIKIPEDIA_GT_1 EN_WIKIPEDIA_GT_1_COUNTS
+do
+  docker-compose exec kafka1 bash -c 'export KAFKA_LOG4J_OPTS="-Dlog4j.rootLogger=DEBUG,stdout -Dlog4j.logger.kafka=DEBUG,stdout" && kafka-topics \
```
mikebin, line 168: Are these log4j options needed?

scripts/start.sh
```diff
@@ -48,6 +48,24 @@ sleep 5
 echo "Creating role bindings for principals"
 docker-compose exec tools bash -c "/tmp/helper/create-role-bindings.sh" || exit 1
 
+# Workaround for setting min ISR on topic _confluent-metadata-auth
+docker-compose exec kafka1 kafka-configs \
```
mikebin, line 52: Some of these CLI commands use port 12091, and some use 11091. Is that intentional?

mikebin (commented):
LGTM, just a couple of minor questions.

scripts/start.sh
ybyzek: @mikebin Without the log4j options, the output includes `“WARN The configuration ... was supplied but isn’t a known config. ”` so it's not required per se, but it generates cleaner output.  WDYT?

scripts/start.sh
ybyzek: The reasoning was to differentiate:

1. the examples using 11091 are "best practice" and require security configs, whereas
2. the examples using 12091 are to the PLAINTEXT port and are workarounds

scripts/start.sh
mikebin: Sounds fine to me. I'm curious why setting `DEBUG` _reduces_ the amount of logging instead of increasing it, though.

scripts/start.sh
mikebin: Makes sense to me.

rspurgeon (approved):
LGTM

PR #249, opened 2020-09-25
DEVX-2124: Catch condition: KSQL is not yet ready to serve requests.

ybyzek:


rspurgeon (approved):
lgtm

scripts/helper/functions.sh
```diff
@@ -80,6 +80,15 @@ host_check_mds_up()
   return 0
 }
 
+host_check_ksqlDBserver_up()
+{
+  KSQLDB_CLUSTER_ID=$(curl -s -u ksqlDBUser:ksqlDBUser http://localhost:8088/info | jq -r ".KsqlServerInfo.ksqlServiceId")
+  if [ "$KSQLDB_CLUSTER_ID" == "ksql-cluster" ]; then
```
colinhicks, line 86: Not sure if this is relevant, but if the user were to customize `ksql.service.id` elsewhere, this would fail even though the server is ready.

I saw `KSQLDB=ksql-cluster` defined elsewhere, although I'm not sure it's in scope here. You could also consider `jq -r 'select(.KsqlServerInfo.ksqlServiceId)'` to detect any truthy value.

colinhicks (approved):
LGTM overall. I noted a minor nit for consideration.

scripts/helper/functions.sh
ybyzek: @colinhicks 

> I saw KSQLDB=ksql-cluster defined elsewhere

Exactly, this is hardcoded when creating the RBAC role bindings.  

Short-term: in a way, we *want* the script to fail on this check here because it's hardcoded elsewhere
Long-term: tracking in DEVX-1876

colinhicks: Ah, that makes sense. Thanks for the context, @ybyzek!

javabrett (approved):
🍰

PR #258, opened 2020-10-07
DEVX-2138: Move demos-common Wikipedia Activity Monitor Kafka Streams…

ybyzek:
… application into cp-demo

This is an almost as-is copy from https://github.com/confluentinc/demos-common to cp-demo in a `kstreams-app` folder. The plan is to decommission demos-common once this PR is merged.

Changes:

1. New files added almost as-is from demos-common, into `kstreams-app` folder.  Only change: `rootProject.name = 'cp-demo-kstreams-app'`
2. Updated `docs/index.rst` with new link to the source code

Validation:

* Code: `(cd kstreams-app && make clean test)`
* Docs: I tried to build locally but it fails because of known errors caused by this week's docs migration.


kstreams-app/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM maven:3.6.1-jdk-8 AS build
```
mikebin, line 1: Is this Dockerfile valid? It references an old CP 5.4.0 version, and files like `pom.xml` which don't seem to exist.

mikebin (commented):
Overall LGTM. Might want to run `gradle clean` and avoid committing `.class` and .`jar` files to github.

kstreams-app/Dockerfile
ybyzek: @mikebin it might be there as an archive?  @rspurgeon may know.

FYI, we are going to rewrite this application using Java Spring, so this PR is meant just as an intermediary step for migration of code into the repo.  So my leaning is to leave everything as-is.

ybyzek:
@mikebin 

> Overall LGTM. Might want to run gradle clean and avoid committing .class and .jar files to github.

Good catch, thank you!  Many files removed, but not sure if I got all of them with `make clean` and `gradle clean`?  One jar is still there (`gradle-wrapper.jar`)

kstreams-app/Dockerfile
rspurgeon: @ybyzek We could update this to use the CONFLUENT config.env parameter now that this is in cp-demo.

@mikebin I'm not sure what you mean "files like pom.xml" which don't seem to exist?

rspurgeon:
@ybyzek could you bring over the 'dotfiles', most importantly the `.gitignore` file (but there is also a `.editorconfig` file back in the original repo

kstreams-app/build.gradle
```diff
@@ -0,0 +1,80 @@
...
+}
+
+group = 'io.confluent'
+version = '0.0.8'
+sourceCompatibility = "1.8"
+targetCompatibility = "1.8"
+
+ext {
+  kafkaVersion = '2.6.0'
+  cpVersion = '6.0.0'
```
rspurgeon, line 23: can we leverage env_files/config.env now that this is in cp-demo?

kstreams-app/Dockerfile
mikebin: @rspurgeon - I've probably just misread, but it looks like the Dockerfile is trying to copy a `pom.xml` to the image, but I don't see a `pom.xml` in the `kstreams-app/` folder - only `build.gradle`.

rspurgeon:
> @mikebin
> 
> > Overall LGTM. Might want to run gradle clean and avoid committing .class and .jar files to github.
> 
> Good catch, thank you! In the latest commit I removed many of those files, but not sure if I got all of them with `make clean` and `gradle clean`? One jar is still there (`gradle-wrapper.jar`)

FYI it is recommended to commit `gradle-wrapper.jar` to VCS: https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper

ybyzek:
@rspurgeon 

> @ybyzek could you bring over the 'dotfiles', most importantly the .gitignore file (but there is also a .editorconfig file back in the original repo

Done with latest commit.

kstreams-app/build.gradle
ybyzek: WDYT https://github.com/confluentinc/cp-demo/pull/258/commits/be73b0a20a04be3a8202c7e354afb80e5e691006 ?

kstreams-app/Dockerfile
ybyzek: @rspurgeon IIUC @mikebin is referring to this line: https://github.com/confluentinc/cp-demo/blob/DEVX-2138/kstreams-app/Dockerfile#L5

And I don't know the history, but my guess is that this application first had a pom.xml and relied on Maven, and then was converted to Gradle...but the reference to the pom file remained.

rspurgeon: :homer-doh:! yes of course.

I think I remember.  At first I used this mvn and this Dockerfile but then I switched to using Jib, so the dockerfile is built from with the gradle jib plugin.  I think this Dockerfile can be removed entirely.

ybyzek: 🥇 beauty.  I removed the Dockerfile and verified `make clean package` rebuilds it.

kstreams-app/build.gradle
rspurgeon: Slick!

rspurgeon (approved):
LGTM

PR #264, opened 2020-10-25
DEVX-2201: ERROR Unable to load token keyPair from config token.key.p…

ybyzek:
…ath: cp-demo broken when host user id gt 1000

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2201

Resolve user permissions when host user is greater than 1000 (Docker containers run as appuser 1000)

### Author Validation

- [x] Documentation: run through a few manual steps in the playbook
- [x] Run cp-demo: demo runs end-to-end when run as a new user with ID 1001 (can repro both failure without PR and fix with PR)

### Reviewer Tasks

- [ ] Documentation
- [ ] Run cp-demo

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -28,6 +28,10 @@ mkdir -p ${DIR}/security/keypair
 openssl genrsa -out ${DIR}/security/keypair/keypair.pem 2048
 openssl rsa -in ${DIR}/security/keypair/keypair.pem -outform PEM -pubout -out ${DIR}/security/keypair/public.pem
 
+# Enable Docker appuser to read files if created by a different UID
+chmod 644 ${DIR}/security/keypair/keypair.pem
+chmod 644 ${DIR}/security/*.key
+
 # Bring up openldap
 docker-compose up -d openldap
 sleep 5
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -28,6 +28,7 @@ mkdir -p ${DIR}/security/keypair
 openssl genrsa -out ${DIR}/security/keypair/keypair.pem 2048
 openssl rsa -in ${DIR}/security/keypair/keypair.pem -outform PEM -pubout -out ${DIR}/security/keypair/public.pem
 
+# Do not make keys world-readable in production
 # Enable Docker appuser to read files if created by a different UID
 chmod 644 ${DIR}/security/keypair/keypair.pem
 chmod 644 ${DIR}/security/*.key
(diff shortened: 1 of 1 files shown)


mikebin (approved):
LGTM. Generally, it's not recommended to have private keys be world-readable, but perhaps not important for a demo like this.

ybyzek:
> Generally, it's not recommended to have private keys be world-readable, but perhaps not important for a demo like this.

@mikebin agree.  I added an extra comment to that effect, but it leaves question how users would do this in production.  This is still an open item for discussion, hope to resolve in Slack perhaps?

rspurgeon:
Another possible solution is to make the `user` of the container match the running host user:group, for example:

In the `kafka1` docker-compose.yml service stanza you could add:

```
user: ${USER_ID}:${GROUP_ID}
```

And in `scripts/start.sh` add: 
```
export USER_ID=$(id -u)
export GROUP_ID=$(id -g)
```

This will set the uid of the container's `appuser` to be match the host uid and the same file permissions will apply.

ybyzek:
@rspurgeon I had effectively tried that with `user: 1001:1001` but it failed due to:

```
kafka1            | ===> Check if /var/lib/kafka/data is writable ...
kafka1            | Command [/usr/local/bin/dub path /var/lib/kafka/data writable] FAILED !
```

rspurgeon:
> @rspurgeon I had effectively tried that with `user: 1001:1001` but it failed due to:
> 
> ```
> kafka1            | ===> Check if /var/lib/kafka/data is writable ...
> kafka1            | Command [/usr/local/bin/dub path /var/lib/kafka/data writable] FAILED !
> ```

I see, the user used during Docker build I guess will cause these issues.

rspurgeon (approved):
I agree that another solution to this is preferrable, however, to ublock LGTM until something else is found

scripts/start.sh
```diff
@@ -28,6 +28,11 @@ mkdir -p ${DIR}/security/keypair
 openssl genrsa -out ${DIR}/security/keypair/keypair.pem 2048
 openssl rsa -in ${DIR}/security/keypair/keypair.pem -outform PEM -pubout -out ${DIR}/security/keypair/public.pem
 
+# Do not make keys world-readable in production
```
andrewegel, line 31:
```suggestion
echo "SETTING INSECURE PERMISSIONS ON '${DIR}/security' FOR DEMO PURPOSES ONLY NEVER USE THIS IN PRODUCTION"
# Do not make keys world-readable in production
```

andrewegel (approved):
I left a suggestion (not required) to print out a nasty message about whats being done here.

scripts/start.sh
ybyzek: @andrewegel good idea to echo this out.  I wordsmithed it and committed in latest commit.

PR #271, opened 2020-11-02
DEVX-2231: add CONTROL_CENTER_STREAMS_CPREST_URL for Self Balancing Clusters

javabrett:
### Description

This PR creates Control Center config property `confluent.controlcenter.streams.cprest.url` which is required for the Self Balancing Clusters UI.

This resolves a 403 unauthorized error from the C3 UI when accessing `/api/kafka-rest`.  With this property set, the Self Balancing Cluster feature is now enabled in the UI.

Note that this change doesn't perform any set-up for demo of the Self Balancing Cluster feature.

### Author Validation

[ ] Documentation
[x] Run cp-demo

### Reviewer Tasks


ybyzek (approved):
LGTM thanks @javabrett

PR #273, opened 2020-11-02
DEVX-2232: Skip cert & image regeneration on subsequent runs (unless CLEAN=true)

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2232

For subsequent cp-demo runs, reuse certs and Connect image.  Result: speed up demo startup by ~30s

### Author Validation

- [x] Documentation
- [x] Run cp-demo

### Reviewer Tasks

- [ ] Run cp-demo


mikebin (approved):
Yay! LGTM

javabrett (approved):
Is there a loose plan to use `FAST=true` in the future to trigger other time-saving, but perhaps not 100%-reset (or other reason to run full/slow), optimisations?

Edit: (following was mostly targeted at cert-generation, but might also apply to image-regen).

I need to do some testing, but now that I see this it is an obviously-good optimisation (thanks!) ... so ... should we just invert this and make it the default, and add a `--clean` or other clean/full/slow "build" (start) option?  Or this could be treated as experimental and flip to the default behaviour later.

Given how hard it can be to work with self-signed certs in the browser especially, and, unless you are specifically working on a change to the cert-generation, it is difficult to imagine that you'd want cert-gen on each run, I can imagine wanting this flipped to be the default, without specifying `FAST=true`.

docs/index.rst
```diff
@@ -94,12 +94,20 @@ Start Example
       cd cp-demo
       git checkout |release_post_branch|
 
-#. From the ``cp-demo`` directory, start the entire example by running a single command that generates the keys and certificates, brings up the Docker containers, and configures and validates the environment. This takes approximately 10 minutes to complete.
+#. From the ``cp-demo`` directory, start the entire example by running a single command that generates the keys and certificates, brings up the Docker containers, and configures and validates the environment.
 
-   .. sourcecode:: bash
+   - The first time you run ``cp-demo``, the startup takes approximately 7 minutes to complete.
```
javabrett, line 99: Re text that describes time-expectations ... since these will vary constantly over time and be subject to optimisations, and may be hardware-dependant ... this change is fine for now, but if we are back here again, maybe we can come up with something a little vaguer ...

> the startup time will depend on your hardware and whether Docker images are already cached on your machine, but expect startup to take more than 5 minutes.

ybyzek: @javabrett Text is now vaguer

ybyzek:
> I can imagine wanting this flipped to be the default, without specifying FAST=true.

@javabrett one reason to not make `FAST=true` is to make the default path the most conservative.  However, this FAST flip crossed my mind as well (specifically FAST default=true, user has to set it to FALSE to go through whole process) so please let me know if you have that leaning (at the expense of the cp-demo run being conservative) and I will flip :)

javabrett:
I use cp-demo like demo-source, so I expect it to have some sort of clean->build->deploy/launch->test/use->... cycle.  So I'd be comfortable with the notion that some parts of a first/clean build might be cached or retained, and that a "clean" might be needed, should the inputs or scripting for those artifacts change ... that is, assuming that the build scripting can't auto-detect a changed "source" and force regeneration of the target.

There's no `pom.xml`, `build.gradle` or `Makefile`, otherwise a `make clean` could force a reset.  Sometimes I use `git clean -dxf` (with care) to force-clean the project.  I notice there is a `scripts/security/certs-clean.sh` but it does assume it is run from its own directory for correct cleanup.

cp-demo currently does a full/clean build for each `start.sh` and it's tricky to give that up, but for things like cert-generation the Connect image build that feels ok.  I should look at the local Docker build ... sometimes with careful source placement for the Docker context, these can cache well.

I don't think I would be that surprised if cert-generation is by-default a one-hit until-cleaned ... given the CA is self-signed, so you have to either import/trust the CA or accept browser-warnings, I think it's better to use the same CA/certs for a longer test-session.  Can always delete them to force regen.

ybyzek:
@javabrett latest commit results in behavior that cert gen & Docker image gen is only on the first run, but note I flipped the logic using a `CLEAN` param instead of `FAST`.

Also includes 2 minor fixes:

- Addresses `I notice there is a scripts/security/certs-clean.sh but it does assume it is run from its own directory for correct cleanup.`
- https://github.com/confluentinc/cp-demo/issues/274 (bad hygiene to include in this PR, but just for ease)

Separate conversation: we can look at more polished solutions using `make clean` (if necessary)

ybyzek:
@javabrett @rspurgeon if you have the availability, could you please review the UX on subsequent cp-demo runs, since the PR flipped from explicit `FAST=true` to default `CLEAN=false`.  If you don't have time to review, no worries.  I did do the validation of the bits.

javabrett:
@ybyzek running now.

scripts/security/certs-clean.sh
```diff
@@ -6,4 +6,5 @@ set -o nounset \
 #    -o xtrace
 
 # Cleanup files
-rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
javabrett, line 9: Above: `    -o verbose`

Now that this script is called by normal workflow, the `verbose` makes it a little chattier than other areas, since it echos raw commands:

```
$ scripts/start.sh
CLEAN=true -> deleting existing certificates and local Connect Docker image generated by cp-demo
#    -o xtrace

# Cleanup files
 cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd
 dirname "${BASH_SOURCE[0]}"
(cd "${DIR}" && rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12)
Untagged: localbuild/connect:6.0.0-6.0.0
```

So for consideration - commenting-out `verbose` per `xtrace` if this also looks a bit chatty/messy output.

scripts/helper/functions.sh
```diff
@@ -62,6 +62,39 @@ get_kafka_cluster_id_from_container()
   return 0
 }
 
+clean_demo_env()
+{
+  local DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+  echo "CLEAN=true -> deleting existing certificates and local Connect Docker image generated by cp-demo"
```
javabrett, line 69: Logic OK, but the output can be confusing.  If I start with a squeaky-clean working copy (no certs) and run:

```
$ CLEAN=false scripts/start.sh
CLEAN=true -> deleting existing certificates and local Connect Docker image generated by cp-demo
#    -o xtrace

# Cleanup files
 cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd
 dirname "${BASH_SOURCE[0]}"
(cd "${DIR}" && rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12)
```

So I asked for `CLEAN=false`, which can't happen because already clean, and the output is immediately `CLEAN=true -> deleting existing certificates ...`, which contradicts, and claims to be deleting existing certificates, but there are none.

The same message appears on a clean repo without the `CLEAN=false` - less contradictory, but still claims to be deleting existing certificates, and there are none.

So the logic is OK, but I think we'll need to do more explaining of the logic in the output here.

javabrett (changes requested):
I plan to add some more UX/output suggestions later.

scripts/helper/functions.sh
ybyzek: 👍 https://github.com/confluentinc/cp-demo/pull/273/commits/48c658b5f4cb1c9f405e7ba006ecd7fa5fe05dee improves UX

scripts/security/certs-clean.sh
ybyzek: 👍 verbose is commented out

scripts/helper/functions.sh
javabrett: ⭐ 🍰 ⭐

javabrett (approved):
Superb - this shaves more time off redeploys.  🍰

PR #277, opened 2020-11-08
Added a basic .dockerignore

javabrett:
### Description 

Connect has a local `docker build` these days - we should have a basic `.dockerignore` to reduce the size of the build-context and avoid any accidents.

None.

### Author Validation

### Reviewer Tasks


ybyzek (approved):
@javabrett LGTM

PR #278, opened 2020-11-09
C3 and ksqldb server certs, change CN and SAN to match hostname

javabrett:
### Description 

This PR will change the CN and SAN on certificates for the specific hostnames for Control Center and ksqlDB Server.  These certificates are currently created with:

* CN=`controlcenter`, SANs=`localhost`, `controlcenter`
* CN=`ksqlDBserver`, SANs=`localhost`, `ksqlDBserver`

This PR changes this to:

* CN=`control-center`, SANs=`localhost`, `control-center`
* CN=`ksqldb-server`, SANs=`localhost`, `ksqldb-server`

Note that this matches the hostnames on the servers, as assigned by `docker-compose.yml` and Docker.  So if you do happen to access these servers via these hostnames, _and_ you are trusting the right truststore, then you will get a matching hostname and a valid cert.

_What behavior does this PR change, and why?_

Not a huge amount, but it does make the certificates more correct.  The reason is that these two servers don't currently communicate with each-other over TLS/HTTPS, and external access is mapped to `localhost` on the Docker host.  It does provide some tidy-up preparation and consistency in preparation for enabling ksqlDB Server TLS/HTTPS.

Prior to this change, say you exec shell into the Control Center host, and run:

```
$ hostname
control-center

$ curl -v --cacert /etc/kafka/secrets/snakeoil-ca-1.crt  https://control-center:9022
* Rebuilt URL to: https://control-center:9022/
*   Trying 172.31.0.9...
* TCP_NODELAY set
* Connected to control-center (172.31.0.9) port 9022 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/kafka/secrets/snakeoil-ca-1.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, [no content] (0):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
[...]

PR #279, opened 2020-11-09
Enable (inactive) TLS/HTTPS for ksqlDB Server

javabrett:
### Description 

This PR:
- Activates port `8089` on ksqlDBServer and maps to the Docker host on same port for HTTPS/TLS access to ksqlDB Server REST.
- Stops generating a unique certificate for `ksqlDBServer`, instead copying the Control Center certificate.  This means that once the user accessing https://localhost:9092 has accepted the self-signed cert, the Websockets `wss://` access will also work.  ~Note that the CN/SAN on the certs is currently off ... this should be re-aligned after #278 is merged, or if it is merged first, can resolve conflicts in this PR.~ **Edit:** #278 was merged into this PR.
- ~Adds a commented-out config option for `CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL` to use the new HTTPS port, mapped to `localhost:8089` on the Docker host.~.  **Edit:** this override now occurs via environment, e.g. `export CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL=https://localhost:8089`.

To test this change, first use http://localhost:9021 as normal, and verify that you can run the query `select * from WIKIPEDIA EMIT CHANGES;` in the ksqlDB UI.  Then, access via https://localhost:9022 - same will not work, as Control Center will try and send `wss://` to port `8088`.  Thirdly, make the config change and redeploy, and access via https://localhost:9022 , then the ksqlDB query UI will work.

Nothing much by-default, mainly prep/enable testing for full HTTPS/TLS mode.  If the new option for `CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL` is uncommented prior to deployment, the when Control Center is accessed over https, https://localhost:9022, and the self-signed certificate is accepted, then Web Sockets `wss://` to the ksqlDB Server container will use the HTTPS/TLS port.

Note that mixed-mode is not currently supported, and Control Center only supports one or the other http or https to ksqlDB Server from the UI, so this is a deploy-time choice currently.

**Edit**

Since #278 was merged with this PR ... the additional change is that the DNS SANs on the shared `control-center` and `ksqldb-server` HTTPS certificates now contain: `localhost`, `control-center` and `ksqldb-server`, i.e. adds `ksqldb-server` and `controlcenter`->`control-center`.  These servers are mostly referred to via `localhost`, but it's nice for them to have certs that match their Docker hostnames anyway, in case you do want to access them and don't want to get non-matching hostname errors.

### Author Validation

- [ ] Documentation: @javabrett  please see staging docs linked from https://github.com/confluentinc/docs-platform/pull/161
- [X] Run cp-demo

### Reviewer Tasks


docker-compose.yml
```diff
@@ -657,6 +657,10 @@ services:
       CONTROL_CENTER_KSQL_WIKIPEDIA_URL: http://ksqldb-server:8088
       # Access ksqlDB server from a browser on your localhost
       CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL: http://localhost:8088
+      # Uncomment following to enable ksqlDB browser access over TLS/HTTPS
```
ybyzek, line 660: @javabrett this is an excellent addition, thank you.  

1. If we keep this option as a comment, please consider adding verbiage that this replaces the line above `CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL: http://localhost:8088`, to clarify that it is a replacement, not addition.  i.e., user should comment above and uncomment this.  Technically, this is not needed (the 2nd line should have precedence) but it would be helpful to call it out to avoid copy/paste errors and to educate users.

2. Another option we should consider is to go beyond just making a comment, but to add a flag (e.g. something like `ENABLE-KSQLDB-TLS`  when running cp-demo, and created a docker-cmpose overrides file with this parameter.  It's more upfront investment but (a) easier to run and (b) easier to document. I lean towards this option because of its advantages.

scripts/security/certs-create.sh
```diff
@@ -11,7 +11,12 @@ rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12 *.log
 # Generate CA key
 openssl req -new -x509 -keyout snakeoil-ca-1.key -out snakeoil-ca-1.crt -days 365 -subj '/CN=ca1.test.confluentdemo.io/OU=TEST/O=CONFLUENT/L=PaloAlto/ST=Ca/C=US' -passin pass:confluent -passout pass:confluent
 
-users=(kafka1 kafka2 client schemaregistry restproxy connect connectorSA controlcenter ksqlDBserver ksqlDBUser appSA badapp clientListen zookeeper mds)
+# we no longer generating a cert for ksqlDB Server - it shares a cert with Control Center
```
ybyzek, line 14: 1. Let's clean up the comment, e.g. `ksqlDB (ksqldbserver) and Control Center (controlcenter) share a certificate`.

2. Can you add to the comment to explain *why* they can't have distinct certificates in this demo?  This will help us remember in 1 year ;)

ybyzek (changes requested):
@javabrett wrt

> this should be re-aligned after #278 is merged

May I suggest that this PR be merged with #278 , and then we can review a single PR?

javabrett:
> @javabrett wrt
> 
> > this should be re-aligned after #278 is merged
> 
> May I suggest that this PR be merged with #278 -- so that there is a working PR with both changes -- and then we can review a single PR?

Closed #278 and will bring those changes into this PR.

docker-compose.yml
javabrett: @ybyzek agreed.  Config for this is now externalised to launch environment via `CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL`, which will default to `http://localhost:8088`.  This comment is left, but we may want to add something in the docs also once reviewed:

```
# To use Control Center and its ksqlDB interface over https, before launching, run in the launch shell:
# export CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL=https://localhost:8089
```

scripts/security/certs-verify.sh
```diff
@@ -5,7 +5,7 @@ set -o nounset \
     -o verbose
 
 # See what is in each keystore and truststore
-for i in kafka1 kafka2 client schemaregistry restproxy connect connectorSA controlcenter ksqlDBserver ksqlDBUser appSA badapp clientListen zookeeper mds
+for i in kafka1 kafka2 client schemaregistry restproxy connect connectorSA control-center ksqldb-server ksqlDBUser appSA badapp clientListen zookeeper mds
```
ybyzek, line 8: @javabrett Can you please elaborate why this PR adds dashes to `control-center ksqldb-server` but not, for example, `ksqlDBUser schemaregistry restproxy`?

scripts/security/certs-create.sh
```diff
@@ -11,7 +11,16 @@ rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12 *.log
 # Generate CA key
 openssl req -new -x509 -keyout snakeoil-ca-1.key -out snakeoil-ca-1.crt -days 365 -subj '/CN=ca1.test.confluentdemo.io/OU=TEST/O=CONFLUENT/L=PaloAlto/ST=Ca/C=US' -passin pass:confluent -passout pass:confluent
 
-users=(kafka1 kafka2 client schemaregistry restproxy connect connectorSA controlcenter ksqlDBserver ksqlDBUser appSA badapp clientListen zookeeper mds)
+# ksqlDB Server (ksqldb-server) and Control Center (control-center) share a certificate; a separate certificate is not generated for ksqldb-server
+# this shared certificate allows the self-signed certificate to be accepted by a Control Center browser user visiting https://localhost:9092 ,
```
ybyzek, line 15: `allows the self-signed certificate` --> just to be explicit, please elaborate _whose_ self-signed certificate?

scripts/security/certs-create.sh
```diff
@@ -11,7 +11,16 @@ rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12 *.log
...
+# without importing and trusting the self-signed CA, and this acceptance will also apply later to Websockets requests to wss://localhost:8089
+# (port-forwarded to ksqldb-server:8089).  This is necessary as browsers never prompt to trust certificates for this kind of wss:// connection,
+# see https://stackoverflow.com/a/23036270/452210 .
```
ybyzek, line 18: Thanks for adding the StackOverflow link, very helpful to remember the caveat & workaround!

scripts/security/certs-create.sh
```diff
@@ -11,7 +11,16 @@ rm -f *.crt *.csr *_creds *.jks *.srl *.key *.pem *.der *.p12 *.log
...
+users=(kafka1 kafka2 client schemaregistry restproxy connect connectorSA control-center ksqlDBUser appSA badapp clientListen zookeeper mds)
```
ybyzek, line 19: > ksqlDB Server (ksqldb-server) and Control Center (control-center) share a certificate

However, the certificate is called just control-center.  This is a touch confusing in the singular name, which could be alleviated if it is named to reflect its dual purpose.  Let's rename `control-center` to something that encompasses both names, e.g. `control-center-and-ksqldb-server`...WDYT?

ybyzek (changes requested):
@javabrett 

Going to https://localhost:9022 results in a Chrome browser warning in the screenshot below.  As a result, can this PR augment docs with a new section on the https port?

1. Mention that this C3 HTTPS 9022 port exists
2. Need to run start script with `export CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL=https://localhost:8089`
3. The expected browser warning
4. How to proceed in the browser
 
Screen Shot 2020-11-13 at 10 25 56 AM

docker-compose.yml
```diff
@@ -656,13 +656,15 @@ services:
       # ksqlDB
       CONTROL_CENTER_KSQL_WIKIPEDIA_URL: http://ksqldb-server:8088
       # Access ksqlDB server from a browser on your localhost
-      CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL: http://localhost:8088
+      # To use Control Center and its ksqlDB interface over https, before launching, run in the launch shell:
```
mikebin, line 659: Will this option be documented anywhere else besides in a comment here?

ybyzek: @mikebin Agree, IMO the whole flow to C3 HTTPS warrants a separate section in docs.  @javabrett will you do this? Or  I'm happy to assist with the docs part if you prefer.

scripts/security/certs-create.sh
javabrett: Tried to improve this comment ...

scripts/security/certs-create.sh
javabrett: I can go either way with the naming - there's something nice about the consistency we are approaching, where each server requiring a cert has one with a filename and a DNS SAN matching its Docker hostname - but existing `mds` and now `control-center` and `ksqldb-server` have SAN script customisations, and now the latter pair are sharing files.  Happy to try with a shared filename.

Pushing two commits - the first just changes to a shared file-name e.g. `kafka.control-center-and-ksqldb-server.[truststore|keystore].jks`, replacing the per-service copies.

The second commit also replaces `control-center` with the common name `control-center-and-ksqldb-server` in the `users` generator list ... this necessitates more specific handling in the DNS SAN processing so that `control-center` can be retained as a valid name, since `$i` gets set to the CN and first DNS SAN.  As it stands at that commit, the common cert will have:

- `CN=control-center-and-ksqldb-server`, which won't match anything, but CN is less important these days
- DNS SAN: `[control-center-and-ksqldb-server, localhost, control-center, ksqldb-server]` ... again `control-center-and-ksqldb-server` is redundant and would require more script logic to remove it, but it's harmless.

@ybyzek please review and let me know what looks better to you.

scripts/security/certs-verify.sh
javabrett: Only these two as a combination of a) these were the only two server certs I was working with in the context of getting the WebSockets working for HTTPS per this PR, and b) they are the only hosts I noticed where their Docker hostnames did not match their name in the certificates iterator, therefore they had non-matching CN/DNS SANs.

Note that this is unimportant in the context of this PR, since access is via port-mapped `localhost` Docker host, so as long as these certs have `localhost`, that part works fine, hence the name-fixes were originally in PR #278 but now included here.

Checking the others:
- [x] `restproxy` -> `hostname: restproxy`
- [x] `schemaregistry` -> inconsistent in `docker-compose.yml` as `hostname:` directive missing, but defaults to container name `schemaregistry`, so correct.
- [x] `ksqlDBUser` -> appears to be a client cert?  Name probably doesn't matter.

Now the real question ... since `control-center` has a hyphenated Docker hostname, why not `rest-proxy` and `schema-registry` 😸 ?

docker-compose.yml
javabrett: I'm happy to contribute to docs - either in a separate PR or this one, but I won't be able to progress docs much this week.

javabrett:
Testing note: since certificate names change based on the new commits, be sure to scrub your security (certs) before testing, since we now have skip-regeneration logic.

scripts/security/certs-verify.sh
ybyzek: @javabrett  to make sure we are in sync, should the desired end state be that they all have hyphens?  If you agree, I can file a separate jira to change RP and SR to `rest-proxy` and `schema-registry` everywhere

scripts/security/certs-create.sh
ybyzek: @javabrett  In this case, I think the ability to reuse the cert generation code is more beneficial than trying to create unique code paths for `control-center-and-ksqldb-server` (and `mds`, for that matter), so the code change LGTM.

I am in the process of running cp-demo with the latest commit at https://github.com/confluentinc/cp-demo/pull/279/commits/eb8cf3b114bcc289ebcb7e519cd2065fc9da58ed, and will report back if there is any friction.

docker-compose.yml
ybyzek: @javabrett unless you're pressed to get the functionality, let's keep this PR open till we get docs added in as well.  I'm happy to assist with the docs piece if it helps.

ybyzek:
@jeanlouisboudart @waliaabhishek -- I wanted to draw your attention to this PR because it slightly modifies cert generation.  I don't think there is impact to https://github.com/confluentinc/jmx-monitoring-stacks but you can assess.

ybyzek:
@javabrett PR at https://github.com/confluentinc/cp-demo/pull/282 improves the UX on selecting HTTPS vs HTTP mode -- please review and let me know what you think.

Also, in case you weren't aware, there's a touch more friction using Safari compared to Chrome (in HTTPS mode).  Should cp-demo be opinionated and tell users to use Chrome?

ybyzek:
@javabrett no rush, but wanted to ask if you were planning on adding docs into this PR or you'd like me to ?

javabrett:
> @javabrett no rush, but wanted to ask if you were planning on adding docs into this PR or you'd like me to ?

Thanks for the ping - this is getting close to the top of my pile ... I'll post something soonish.

javabrett:
@ybyzek Added some docs.

ybyzek:
@javabrett please see https://github.com/confluentinc/cp-demo/pull/301 --> docs tweaks, validated local build.  Once you review/merge that in, I think this PR will be ready to go ;)

ybyzek (approved):
@javabrett LGTM.

Note: this PR will potentially resolve in `pint merge` conflicts due to changes in https://github.com/confluentinc/cp-demo/pull/292 .  Please take into consideration when ordering PR merges...maybe even rebase one of them before merge.

PR #282, opened 2020-11-18
UX improvement on selecting HTTPS versus HTTP mode

ybyzek:
### Description 

Pull logic into the start script, plus minor tweaks

### Author Validation

### Reviewer Tasks


mikebin (approved):
LGTM

scripts/start.sh
```diff
@@ -14,6 +14,20 @@ export REPOSITORY=${REPOSITORY:-confluentinc}
...
 export CONNECTOR_VERSION=${CONNECTOR_VERSION:-$CONFLUENT}
 
+# Control Center and ksqlDB server must both be HTTP or both be HTTPS; mixed modes are not supported
+# C3_KSQLDB_HTTPS=false: set Control Center and ksqlDB server to use HTTP (default)
+# C3_KSQLDB_HTTPS=true : set Control Center and ksqlDB server to use HTTPS
+C3_KSQLDB_HTTPS=${C3_KSQLDB_HTTPS:-false}
+if [[ "$C3_KSQLDB_HTTPS" == "false" ]]; then
+  export CONTROL_CENTER_KSQL_WIKIPEDIA_URL="http://ksqldb-server:8088"
+  export CONTROL_CENTER_KSQL_WIKIPEDIA_ADVERTISED_URL="http://localhost:8088"
+  C3URL=http://localhost:9021
```
javabrett, line 24: All other `URL` variables are preceded by an underscore, consider `C3_URL` here.

scripts/start.sh
```diff
@@ -14,6 +14,20 @@ export REPOSITORY=${REPOSITORY:-confluentinc}
 # and expects user to build and provide a local file confluentinc-kafka-connect-replicator-${CONNECTOR_VERSION}.zip
 export CONNECTOR_VERSION=${CONNECTOR_VERSION:-$CONFLUENT}
 
+# Control Center and ksqlDB server must both be HTTP or both be HTTPS; mixed modes are not supported
+# C3_KSQLDB_HTTPS=false: set Control Center and ksqlDB server to use HTTP (default)
+# C3_KSQLDB_HTTPS=true : set Control Center and ksqlDB server to use HTTPS
+C3_KSQLDB_HTTPS=${C3_KSQLDB_HTTPS:-false}
+if [[ "$C3_KSQLDB_HTTPS" == "false" ]]; then
+  export CONTROL_CENTER_KSQL_WIKIPEDIA_URL="http://ksqldb-server:8088"
```
javabrett, line 22: These two would prevent me from setting these separately in my own environment, should I wish to do that for some reason.  I'm not that fussed, but should we allow pre-set via `:-` here and following line?

javabrett (approved):
Nice - simplifies enabling HTTPS without needing to know the proper settings for ksqlDB server.

scripts/start.sh
ybyzek: Ack and fixed

scripts/start.sh
javabrett: I guess the problem here is that this variable has to be exported (does it?) so wouldn't get reset ... ignore this I think.

ybyzek: Ack and resolved

PR #287, opened 2020-11-25
Fix struct access syntax in ksqlDB query

theturtle32:
### Description 

For whatever reason, one of the ksqlDB queries bundled with this demo did not use the proper syntax for accessing a struct property. The result was that part of the demo was broken. The query that creates the `en_wikipedia_gt_1` table never resulted in any data with the old syntax.

According to the [official documentation](https://docs.ksqldb.io/en/latest/how-to-guides/query-structured-data/) the proper syntax is `field->sub_field` not `'field.sub_field'`. I'm not sure why this other syntax is here in the example? Maybe the ksqlDB syntax changed and this example was not updated? I'm new to ksqlDB, and came across this as part of the process of evaluating it as a potential technology to incorporate into our operations.

### Author Validation

With the correction, the `en_wikipedia_gt_1` table and corresponding Kafka topic get populated with the expected data. Before the correction, the topic and table remained empty no matter how long the demo application ran.

- [x] Run cp-demo

### Reviewer Tasks

- [ ] Run cp-demo


ybyzek (approved):
@theturtle32 great find, thanks for the PR.

PR #283, opened 2020-11-19
Switch to a cipher suite that Chrome supports.

theturtle32:
### Description 

_What behavior does this PR change, and why?_

The SSL Cipher Suite specified in the docker-compose.yml file is not supported by Google Chrome, as it is apparently considered to be insecure. This PR changes the cipher suite to a similar one that is supported by Google Chrome to allow for connecting to e.g. the Schema Registry server from the browser for manual data exploration via the REST API.

### Author Validation

I have run the demo on Docker on Linux using the `scripts/start.sh` script and it continues to work fine after this change, with the notable difference that my Chrome browser will now allow me to make TLS connections to e.g. the schema registry on port 8085.

- [x] Run cp-demo

### Reviewer Tasks

Accept this if it seems reasonable to you. It's a pretty low-risk, low-impact configuration change.

- [ ] Run cp-demo

PR #292, opened 2020-11-29
DEVX-2270/DEVX-2276/DEVX-2284: consistent, strong TLS 1.3+1.2 cipher suites for all services

javabrett:
### Description 

- Control Center: fixes cipher suite variable name, sets consistent/strong cipher suites for TLS 1.3 and 1.2.
- Connect, Schema Registry, REST Proxy: Enables TLS 1.3 retaining 1.2, sets consistent/strong cipher suites for TLS 1.3 and 1.2.
- Brokers: sets consistent/strong cipher suites for TLS 1.3 and 1.2.  Note that server-preference is not available.
- MDS: sets consistent/strong cipher suites for TLS 1.3 and 1.2.
- ZooKeeper: sets consistent/strong cipher suites for TLS 1.2.  Note that TLS 1.3 and server-preference are not available.

_How to test pre/post PR_

```
git checkout https://github.com/drwetter/testssl.sh.git
cd testssl.sh

./testssl.sh https://localhost:9022 > testssl.control-center.9022.log &

./testssl.sh https://localhost:8083 > testssl.connect.8083.log &
./testssl.sh https://localhost:8085 > testssl.schema-registry.8085.log &
./testssl.sh https://localhost:8086 > testssl.rest-proxy.8083.log &

./testssl.sh https://localhost:8091 > testssl.mds1.8091.log &
./testssl.sh https://localhost:8092 > testssl.mds2.8092.log &

./testssl.sh localhost:2182 > testssl.zookeeper.2182.log &

./testssl.sh localhost:10091 > testssl.broker.10091.log &
./testssl.sh localhost:10092 > testssl.broker.10092.log &
./testssl.sh localhost:11091 > testssl.broker.11091.log &
./testssl.sh localhost:11092 > testssl.broker.11092.log &
```

Compare reports pre/post PR.

This PR can be considered following https://github.com/confluentinc/cp-demo/pull/283 .

### Author Validation

- [ ] Documentation
- [X] Run cp-demo
- [ ] jmx-monitoring-stacks

### Reviewer Tasks

- [ ] Documentation
- [X] Run cp-demo
- [X] jmx-monitoring-stacks


ybyzek:
1. @javabrett If these are "consistent/strong cipher suites for TLS 1.3 and 1.2", do we know why the values specified in this PR are not the default values in each component already?

2. IIUC the purpose of creating parameter names is to avoid duplication of long values, but using `config.env` in this fashion is overloading that file's purpose, which was to set CP versions.  Instead, can we consider:

(a) avoid parameter names and write out each value in the docker-compose.yml file or 

(b) set these parameters in the start script (e.g. https://github.com/confluentinc/cp-demo/blob/6.0.0-post/scripts/start.sh#L7-L31) or 

(c) create a new file that sets all these parameters plus https://github.com/confluentinc/cp-demo/blob/6.0.0-post/scripts/start.sh#L7-L31 . 

cc: @rspurgeon for additional input

mikebin (commented):
Overall LGTM. I assume ksqlDB was not included in this because the configs don't exist in 6.0.0 for cipher suites and protocol version? It should hopefully be included in 6.0.1: https://github.com/confluentinc/ksql/pull/6558.

javabrett:
> 1. @javabrett If these are "consistent/strong cipher suites for TLS 1.3 and 1.2", do we know why the values specified in this PR are not the default values in each component already?  (i.e., is there a product jira to set these by default?)

For services hosted by Jetty server, if the component's `ssl.cipher.suites` is not set, the consistent default seems to be to use Jetty's default allowed ciphers.  Per https://github.com/eclipse/jetty.project/blob/68e70b47decb3bb27ca83be7131bbb4abd86d9b9/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java#L124 , this will be a filtered set of what the current JDK and its JSSE provider allows, blocking a numner of quite old and well-known "considered insecure" ciphers.

And since this commit https://github.com/eclipse/jetty.project/commit/c5e4abd792e7ed3f5e086d9835b25c5b93afe3d5 , Jetty has been enforcing server-order preference on that subset of ciphers.

For Java non-Jetty services, unless the property is set we will see the full JDK JSSE set of allowed cipher suites, and server-order won't be set.  This is fixed in JDK13+ per https://bugs.openjdk.java.net/browse/JDK-8219657 .

There will be a natural lag and difference between what Jetty or JDK filters-out as "unaccptable" ciphers, due to release-cadences, and differences in security postures for what makes an unacceptable cipher.

I think Confluent Platform components avoid setting default cipher suites and server-preferred orders because a) it would create an additional maintenance point as cipher suites available change over time, and b) it makes assumptions on the out-of-the-box tradeoffs between cipher security and client-compatibility.

For the services which don't currently set cipher suites, the presented suites aren't too bad, there's nothing disasterous in there - partly through JDK maintanence which eliminates bad cipher suites by-default over time.  But there's new CVEs all the time and changing security postures on them - e.g. DH 1024 key-exchanges with weak/common primes are now considered by some as a weakness and that these primes could come under a well-resourced attack, which would compromise many ciphers and servers at once if compromised.  Non-streaming encryption block-ciphers such as CBC have also shown some weakness, and GCM is now preferred, and streaming-ciphers only are offered in TLS 1.3.

So I think it is reasonable for the components to continue to have no default for cipher suites, but for `cp-demo` to be opinionated on what a top-grade set of ciphers is.  Also, since Connect, SR and REST Proxy are accidentally blocking TLS 1.3 via the configured suite, and that suite has lost TLS 1.2 support in user-agents e.g Chrome per #283 (Postman is also affected), we need to add suites, so it is reasonable to ask: which exact set of suites should we add, and can we avoid any TLS 1.2 suites which are on the way out in terms of safety and popularity.

javabrett:
> 2. IIUC the purpose of creating parameter names is to avoid duplication of long values, but using `config.env` in this fashion is overloading that file's purpose, which was to set CP versions.

Yes the goals here were DRY and externalise config values from scripts and the services YAML.  I also have a longer-term goal of externalising all required static config from `start.sh` if that's possible, so that said config can be sourced into the current shell, allowing containers to be started/stopped individually if possible for development, without a full `stop.sh`/`start.sh` cycle.  So I much prefer option c).

Another option, if there's nothing wrong with `.env` Docker Compose environment files other than that the default one is claimed by release-eng for platform versions ... Docker Compose [supports multiple `.env` files](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option), but I haven't used that before, and it might involve it's own repetition.

docker-compose.yml
```diff
@@ -156,6 +158,7 @@ services:
       KAFKA_SSL_KEY_CREDENTIALS: kafka1_sslkey_creds
       KAFKA_SSL_TRUSTSTORE_FILENAME: kafka.kafka1.truststore.jks
       KAFKA_SSL_TRUSTSTORE_CREDENTIALS: kafka1_truststore_creds
+      KAFKA_SSL_CIPHER_SUITES: ${KAFKA_SSL_CIPHER_SUITES}
```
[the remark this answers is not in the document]
javabrett: Thanks @mikebin , I was not aware that config option is pending for ksqlDB.  I left ksqlDB out as it isn't TLS-enabled until #279 is merged.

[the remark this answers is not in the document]
javabrett: > Is TLSv1.3 implied here (and for the other components) because of the cipher suite and/or because it's Java 11?

That's correct here for the broker.  Since we have a fixed JDK11 in-container here, we will get TLS 1.3 if the client supports it, otherwise TLS 1.2.

https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#ssl.protocol
https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#ssl.enabled.protocols

For the Jetty-based components, in the absence of specific config we'll be picking-up the Jetty default, which tests as 1.3 (final) if client supported, else 1.2.

ybyzek:
@javabrett thanks for the detailed explanation (cipher suit availability and compatibility is an area I'm not familiar with).

The changes in this PR generally are very good, especially addressing points like "since Connect, SR and REST Proxy are accidentally blocking TLS 1.3 via the configured suite".  

Also, one more side bar comment/question.  It's good to think about balancing:

1. "cp-demo to be opinionated on what a top-grade set of ciphers is" -- being opinionated is good
2. cp-demo maintainability -- sometimes being opinionated comes with overhead of maintenance (versus, say, using product defaults)

Do you have a sense for maintenance requirements in keeping the configured ciphers up to date in `cp-demo`?  Given that there are some actual _misconfigs_ in cp-demo now, we should still merge this PR, but I'm just trying to anticipate maintenance moving forward.  (That's why I was also asking about product defaults, to improve product UX but also to ease the maintenance here).

docker-compose.yml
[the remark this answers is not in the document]
ybyzek: @mikebin nice catch. Minor: https://github.com/confluentinc/ksql/pull/6559 is the specific PR that backports to 6.0.1.  

@javabrett for cleanliness, it would be nice to make this change once in cp-demo for all components, instead of two parts.  WDYT about retargeting/rebasing this PR to 6.0.1 instead of 6.0.0, to be able to accommodate ksqlDB https://github.com/confluentinc/ksql/pull/6559?  I don't know precisely when 6.0.1 is slated to GA, but if we were to merge this before GA, it would target branch `6.0.x`...or after GA, it would target branch `6.0.1-post`.

ybyzek:
@javabrett  https://github.com/confluentinc/cp-demo/pull/283 has been merged (and I did a `pint merge` call the way through).  So if you decided to rebase for 6.0.1, you are clear.  Please note the timing: GA of 6.0.1 is very soon, and it will affect which branch to rebase to...

docker-compose.yml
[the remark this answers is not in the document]
javabrett: @ybyzek I've created #294 for the ksqlDB server enabled protocols and ciphers, targeting `6.0.x` branch and picking-up the required ksqlDB 6.0.1 fix allowing this configuration.  That PR also requires #279 to be active but is config only and can be merged in-advance.

If 6.0.1 GA beats this PR in then I can rebase and retarget it.

javabrett:
> Do you have a sense for maintenance requirements in keeping the configured ciphers up to date in `cp-demo`? Given that there are some actual _misconfigs_ in cp-demo now, we should still merge this PR, but I'm trying to anticipate maintenance moving forward. (That's why I was also asking about product defaults, to improve product UX but also to ease the maintenance here).

There are a couple of components that could use better defaults, and I am pursuing those.  Some of those improvements will come with new JDK version support.

I would hope that maintenance of these cipher settings in `cp-demo` is a fairly low effort commitment over time and something that won't be too easy to misconfigure.  New versions of TLS are infrequent, and cipher suites don't change too frequently either.  Also any CVEs are usually fairly well read.  Either way in terms of providing a good reference of strong ciphers to use and how to use them, `cp-demo` should be more agile than all the individual components, so I recommend setting these common cipher suites.

ybyzek:
@javabrett latest run (at commit f2d3c597b04ddca3f7ff492e16b4d2db4ad15e80) generates error in my environment:

```
kafka1            | [2020-12-01 14:09:59,888] INFO [ZooKeeperClient Kafka server] Closing. (kafka.zookeeper.ZooKeeperClient)
kafka1            | [2020-12-01 14:10:00,742] INFO Client successfully logged in. (org.apache.zookeeper.Login)
kafka1            | [2020-12-01 14:10:00,742] INFO Client will use DIGEST-MD5 as SASL mechanism. (org.apache.zookeeper.client.ZooKeeperSaslClient)
kafka1            | [2020-12-01 14:10:00,743] INFO Opening socket connection to server zookeeper/172.28.0.3:2182. Will attempt to SASL-authenticate using Login Context section 'Client' (org.apache.zookeeper.ClientCnxn)
kafka1            | [2020-12-01 14:10:00,867] INFO SSL handler added for channel: [id: 0x0f57d8ca] (org.apache.zookeeper.ClientCnxnSocketNetty)
kafka1            | [2020-12-01 14:10:00,870] INFO Socket connection established, initiating session, client: /172.28.0.4:33150, server: zookeeper/172.28.0.3:2182 (org.apache.zookeeper.ClientCnxn)
kafka1            | [2020-12-01 14:10:00,872] INFO channel is connected: [id: 0x0f57d8ca, L:/172.28.0.4:33150 - R:zookeeper/172.28.0.3:2182] (org.apache.zookeeper.ClientCnxnSocketNetty)
kafka1            | [2020-12-01 14:10:00,873] INFO channel is disconnected: [id: 0x0f57d8ca, L:/172.28.0.4:33150 ! R:zookeeper/172.28.0.3:2182] (org.apache.zookeeper.ClientCnxnSocketNetty)
kafka1            | [2020-12-01 14:10:00,873] INFO channel is told closing (org.apache.zookeeper.ClientCnxnSocketNetty)
kafka1            | [2020-12-01 14:10:00,876] INFO channel is told closing (org.apache.zookeeper.ClientCnxnSocketNetty)
kafka1            | [2020-12-01 14:10:00,878] INFO Session: 0x0 closed (org.apache.zookeeper.ZooKeeper)
kafka1            | [2020-12-01 14:10:00,879] INFO EventThread shut down for session: 0x0 (org.apache.zookeeper.ClientCnxn)
kafka1            | [2020-12-01 14:10:00,880] INFO [ZooKeeperClient Kafka server] Closed. (kafka.zookeeper.ZooKeeperClient)
kafka1            | [2020-12-01 14:10:00,884] ERROR Fatal error during KafkaServer startup. Prepare to shutdown (kafka.server.KafkaServer)
kafka1            | kafka.zookeeper.ZooKeeperClientTimeoutException: Timed out waiting for connection while in state: CONNECTING
kafka1            | 	at kafka.zookeeper.ZooKeeperClient.waitUntilConnected(ZooKeeperClient.scala:262)
```

Due to ZooKeeper error:

```
zookeeper         | java.lang.IllegalArgumentException: Unsupported CipherSuite: "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
```

javabrett:
> @javabrett latest run (at commit f2d3c59) generates error in my environment:

Compared `docker-compose` versions, and I am using a newer version which allows new features in `.env` processing via `python-dotenv`.  Specifically, in older versions of Docker Compose, both quotation marks and variable-interpolation in `.env` don't work.

I've pushed two commits to fix that, and re-tested using Docker Compose `1.16.0`, which is the oldest we can support with v2.3 compose file format.

This doesn't yet address the feedback on actually using that same file.

javabrett:
@ybyzek I've pushed another 2 commits ... the first could be in another separate PR, but fits here and addresses the feedback about not further overloading `.env` and `env_files/config.env` - it moves the essential exported environment variables (the ones required in `docker-compose.yml` - from `start.sh` to `env.sh` which is sourced from `start.sh`. In future this might support being able to `source scripts/env.sh` manually in your current shell, so `docker-compose` can be run manually and all the variables are properly set (not tested fully yet).

So now the new SSL/TLS cipher suite variables can be moved to `env.sh` and they can still be made available to scripted or non-scripted docker-compose invocations without too much fuss.

scripts/env.sh
```diff
@@ -0,0 +1,19 @@
...
+#-------------------------------------------------------------------------------
+
+# REPOSITORY - repository (probably) for Docker images
+# The '/' which separates the REPOSITORY from the image name is not required here
+export REPOSITORY=${REPOSITORY:-confluentinc}
+
+# If CONNECTOR_VERSION ~ `SNAPSHOT` then cp-demo uses Dockerfile-local
+# and expects user to build and provide a local file confluentinc-kafka-connect-replicator-${CONNECTOR_VERSION}.zip
+export CONNECTOR_VERSION=${CONNECTOR_VERSION:-$CONFLUENT}
+
```
ybyzek, line 15:
```suggestion

# Set consistent and strong cipher suites across all services
```

docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
       ZOOKEEPER_SSL_TRUSTSTORE_LOCATION: /etc/kafka/secrets/kafka.zookeeper.truststore.jks
       ZOOKEEPER_SSL_TRUSTSTORE_PASSWORD: confluent
       ZOOKEEPER_SSL_TRUSTSTORE_TYPE: JKS
+      ZOOKEEPER_SSL_ENABLED_PROTOCOLS: "TLSv1.2"
```
ybyzek, line 36: If ZK does not support v1.3, why is it required to configure v1.2?  --> Can this line be removed?

scripts/start.sh
```diff
@@ -2,18 +2,10 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 source ${DIR}/helper/functions.sh
-source ${DIR}/../env_files/config.env
+source ${DIR}/env.sh
```
ybyzek, line 5: Now that the other params are removed, perhaps this `CLEAN` logic should be moved to sit immediately prior to line 31:

```
# <<Move lines here>>
if [[ "$CLEAN" == "true" ]] ; then
```

docker-compose.yml
javabrett: For cp-demo yes this can be removed.  I'm trying to recall if there are any JDKs or other factors that for ZooKeeper allow TLS 1.1 or earlier, but if there are they don't apply to cp-demo ... so this setting is only there for completeness, but would block TLS 1.3 should it ever be supported, so will remove.

scripts/start.sh
javabrett: Good idea, done.

ybyzek (approved):
@javabrett validated locally, LGTM.  Very clean PR, great optimization, thank you!

PR #300, opened 2020-12-04
DEVX-2277: Change topic creation to use Confluent Server v3 REST API

javabrett:
### Description 

I was using the v3 REST API with `cp-demo` and figured changing topic-creation to use the v3 REST API would be a good addition.  This PR retires shelling-in to a container to run CLI tool `kafka-topics`, replacing with `POST` calls to https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics .

This appears to be much faster, and would eliminate the need for efforts to parallelise topic-creation per #285 .

`curl` was missing from pre-flight installed checks, so added that too.

### Author Validation

- [ ] Documentation
- [X] Run cp-demo
- [ ] jmx-monitoring-stacks

### Reviewer Tasks

Some questions for reviewer:

- Currently we don't bother to use the same user principal to create the topics as before - they are all created by `superUser`.  Does it matter which user creates the topic?
- Note currently sending `Authorization: Basic` and the base64 credentials - would `curl -u` / `--user` be preferred?

### DIFF
--- a/scripts/helper/create-topics.sh
+++ b/scripts/helper/create-topics.sh
@@ -4,44 +4,46 @@ IFS=$'\n\t'
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 
-KAFKA_CLUSTER_ID=$(curl --insecure --location --request GET 'https://localhost:8091/kafka/v3/clusters' --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' | jq -r '.data[0].cluster_id')
+KAFKA_CLUSTER_ID=$(curl -s --insecure --location --request GET 'https://localhost:8091/kafka/v3/clusters' --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' | jq -r '.data[0].cluster_id')
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+echo "cluster_id: ${KAFKA_CLUSTER_ID}"
+
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name users --arg confluent_value_schema_validation "true" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name users --arg confluent_value_schema_validation "true" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "wikipedia.parsed" --arg confluent_value_schema_validation "true" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "wikipedia.parsed" --arg confluent_value_schema_validation "true" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "wikipedia.parsed.count-by-domain" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "wikipedia.parsed.count-by-domain" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "wikipedia.failed" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "wikipedia.failed" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "WIKIPEDIABOT" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "WIKIPEDIABOT" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "WIKIPEDIANOBOT" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "WIKIPEDIANOBOT" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "EN_WIKIPEDIA_GT_1" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "EN_WIKIPEDIA_GT_1" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq) | jq
 
-curl --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
 --header 'Content-Type: application/json' \
 --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
---data-binary @<(jq -n --arg topic_name "EN_WIKIPEDIA_GT_1_COUNTS" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq)
+--data-binary @<(jq -n --arg topic_name "EN_WIKIPEDIA_GT_1_COUNTS" --arg confluent_value_schema_validation "false" -f ${DIR}/topic.jq) | jq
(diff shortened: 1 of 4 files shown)


ybyzek:
@javabrett absolutely phenomenal idea!  And it showcases REST Proxy.  Will test shortly and leave feedback.

scripts/helper/create-topics.sh
```diff
@@ -1,47 +1,49 @@
...
-done
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+KAFKA_CLUSTER_ID=$(curl -s --insecure --location --request GET 'https://localhost:8091/kafka/v3/clusters' --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' | jq -r '.data[0].cluster_id')
+
+echo "cluster_id: ${KAFKA_CLUSTER_ID}"
+
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
```
ybyzek, line 13: Remove all instances of static credentials `c3VwZXJVc2VyOnN1cGVyVXNlcg==` ... should be `superUser:superUser`?

scripts/helper/create-topics.sh
```diff
@@ -1,47 +1,49 @@
...
+--data-binary @<(jq -n --arg topic_name users --arg confluent_value_schema_validation "true" -f ${DIR}/topic.jq) | jq
+
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
```
ybyzek, line 16: For comparison (and really promoting if these examples can be or should be consistent), the cp-demo docs shows an example as follows:

```
      docker-compose exec restproxy curl -X POST \
         -H "Content-Type: application/json" \
         -H "accept: application/json" \
         -d "{\"topic_name\":\"dev_users\",\"partitions_count\":64,\"replication_factor\":2,\"configs\":[{\"name\":\"cleanup.policy\",\"value\":\"compact\"},{\"name\":\"compression.type\",\"value\":\"gzip\"}]}" \
         --cert /etc/kafka/secrets/mds.certificate.pem \
         --key /etc/kafka/secrets/mds.key \
         --tlsv1.2 \
         --cacert /etc/kafka/secrets/snakeoil-ca-1.crt \
         -u appSA:appSA \
         "https://kafka1:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" | jq
```

There are differences between what the cp-demo docs show and what is written here.  Can we think about if it makes sense to synchronize them (e.g. the user/auth, `--insecure` versus not`, passing in the certs...etc)

scripts/helper/create-topics.sh
```diff
@@ -1,47 +1,49 @@
...
-     --bootstrap-server kafka1:11091 \
-     --command-config /etc/kafka/secrets/ksqlDBUser_without_interceptors_ssl.config \
-     --topic "$t" \
-     --create \
-     --replication-factor 2 \
-     --partitions 2
-done
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+KAFKA_CLUSTER_ID=$(curl -s --insecure --location --request GET 'https://localhost:8091/kafka/v3/clusters' --header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' | jq -r '.data[0].cluster_id')
```
ybyzek, line 7: For your consideration, there's a function that does this `get_kafka_cluster_id_from_container` --> https://github.com/confluentinc/cp-demo/blob/6.0.0-post/scripts/helper/functions.sh#L58-L68 .  It would be nice to pull in the helper scripts and then use that function to get the ID.

ybyzek (changes requested):
@javabrett awesome time improvement (~55s before, 8s after).  Please note requested changes

scripts/helper/functions.sh
```diff
@@ -35,7 +35,7 @@ verify_installed()
 preflight_checks()
 {
   # Verify appropriate tools are installed on host
-  for cmd in jq docker-compose keytool docker openssl xargs awk; do
+  for cmd in curl jq docker-compose keytool docker openssl xargs awk; do
```
ybyzek, line 38: Good addition

scripts/helper/create-topics.sh
```diff
@@ -1,47 +1,49 @@
...
+
+echo "cluster_id: ${KAFKA_CLUSTER_ID}"
+
+curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
```
ybyzek, line 11: To avoid the repetitiveness of the `curl` command with all the arguments, would it make sense to make a function here `create_topic()`?

pseudo code:

```
create_topic() {
  topic_name=$1
  confluent_value_schema_validation=$2

  curl -s --insecure --location --request POST "https://localhost:8091/kafka/v3/clusters/${KAFKA_CLUSTER_ID}/topics" \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic c3VwZXJVc2VyOnN1cGVyVXNlcg==' \
--data-binary @<(jq -n --arg topic_name $topic_name --arg confluent_value_schema_validation "true" -f ${DIR}/topic.jq) | jq
}
```

javabrett:
I am blocked implementing code-review requested changes:
- Existing `get_kafka_cluster_id_from_container` relies on `jq`, is only called from `create-role-bindings.sh` which runs on `tools`.
- I also want to use `jq` both as a templating-engine for rendering the POST payload for create topic, and for pretty-printing results.
- As previously discussed, `tools` is scheduled for deprecation and shouldn't be used further, so a new mechanism will be required for `get_kafka_cluster_id_from_container`.  That might be a new CLI container, but waiting to learn if it will include tools like `jq` in the base image.
- If `tools` is going away without a replacement `jq` in-container, options include a) adding `jq` to custom Connect image since we already build one, b) pull-up `get_kafka_cluster_id_from_container` to run on host since the host already has `curl` and `jq` dependencies, c) create a different `tools` based on existing base containers, d) wait for new CLI container and possible add `jq` to it, e) investigate running on `openldap`, which has `jq` but no `curl`.

scripts/helper/create-topics.sh
javabrett: Done.

scripts/helper/create-topics.sh
javabrett: Done, depends on the `tools` container though.

scripts/helper/create-topics.sh
javabrett: Done.

scripts/helper/create-topics.sh
javabrett: Script code is now mostly consistent with this, remaining differences:
- `-d` data, ours is dynamically generated for each call from a `jq` template
- We aren't sending a `key` ... we rely on Basic auth ... above sets both `key` and `user`, should probably pick one.
- Example sets `--tls1.2` forcing a less-secure-than-available protocol (1.3 is best, available) - doc-bug?

scripts/start.sh
```diff
@@ -77,7 +77,7 @@ docker-compose up -d schemaregistry connect control-center
 
 echo
 echo -e "Create topics in Kafka cluster:"
-docker-compose exec kafka1 bash -c "/tmp/helper/create-topics.sh" || exit 1
+docker-compose exec tools bash -c "/tmp/helper/create-topics.sh" || exit 1
```
ybyzek, line 80: @javabrett can you please remind me, why this is executed on the `tools` container (which might change in the future) instead of `kafka1`?

scripts/helper/functions.sh
```diff
@@ -257,3 +257,31 @@ END
...
+    --header 'Content-Type: application/json' \
+    --header 'Accept: application/json' \
+    --data-binary @<(jq -n --arg topic_name "${topic_name}" --arg confluent_value_schema_validation "${confluent_value_schema_validation}" -f ${DIR}/topic.jq) \
+    "https://${broker_host_port}/kafka/v3/clusters/${cluster_id}/topics") && RC=$? || RC=$?
+
+  echo $RESULT | jq || true
+
+  if [[ $RC -ne 0 || -z $RESULT || $RESULT =~ "error_code" ]]; then
+    echo "ERROR: create topic failed $RESULT"
+    return 1
```
ybyzek, line 283: Should this error code be handled by the calling script, i.e., fail fast?

scripts/helper/create-topics.sh
ybyzek: @javabrett 

> We aren't sending a key ... we rely on Basic auth ... above sets both key and user, should probably pick one.

When removing `key`, the command errors out with: `parse error: Invalid numeric literal at line 1, column 5`.  Does this happen in your env and do you know why?

> Example sets --tls1.2 forcing a less-secure-than-available protocol (1.3 is best, available) - doc-bug?
 
IIRC `v1.2` was specified NOT instead of `v1.3` but instead of `v1.1`.  Is your recommendation to remove the version completely from the command examples in the docs, or to explicitly set `v1.3`?

scripts/start.sh
javabrett: I accidentally resolved a conversation - https://github.com/confluentinc/cp-demo/pull/300#discussion_r546328817 and also see https://github.com/confluentinc/cp-demo/pull/300#issuecomment-748557592 .

The new scripting requires `curl` and `jq`, as does the suggested `get_kafka_cluster_id_from_container`, which also runs on `tools`, since it is the only container that has both.

I think we might be best to defer solving this until we know what is available for the `tools` replacement - i.e. we might need a local utils image, or the new CLI image might provide.

If we can't run on `tools` we would also back-out the use `get_kafka_cluster_id_from_container` refactor, which needs to run on `tools`.

scripts/helper/create-topics.sh
javabrett: > Does this happen in your env and do you know why?

IIUC this is because user `appSA` has only been granted `ResourceOwner` on some topics, so might not have grant to create a topic?  I might be misreading the grants.

Changing to `-u superUser:superUser` with no keys or certs works with Basic auth.  Maybe it's more correct to use the certs anyway?

javabrett: > IIRC v1.2 was specified NOT instead of v1.3 but instead of v1.1. Is your recommendation to remove the version completely from the command examples in the docs, or to explicitly set v1.3?

@ybyzek you are correct here, my error ... `--tlsv1.2` sets a minimum, so TLS v1.3 is typically being negotiated anyway, preventing TLS 1.1 or worse only.

I'll update the PR to include the same guide-rails.

scripts/start.sh
ybyzek: Agree, let's leave it on `tools` for now.

ybyzek:
@javabrett please note PR https://github.com/confluentinc/cp-demo/pull/316 will impact this PR due to topic renaming.  Maybe this PR should retarget 6.1.x to avoid point merge conflicts?

javabrett:
@ybyzek Ack that there will be changes/conflict to resolve here, assuming #316 is merged first.

Will those changes also be required assuming #316 applies also on 6.1.x?  I'm happy to target that branch, but equally happy to stay on 6.0.1-post, unless there are other indications that suggest 6.1.x is better.

scripts/helper/create-topics.sh
javabrett: Added `--tlsv1.2` to `curl` for create topic.

scripts/helper/functions.sh
javabrett: This should already fail-fast I think (and according to a manual test), via:

https://github.com/confluentinc/cp-demo/blob/d0649528f5d1f27e5b6f32ee682881369ac97ad0/scripts/helper/create-topics.sh#L2

If I change:

```
diff --git a/scripts/helper/create-topics.sh b/scripts/helper/create-topics.sh
index 7adfca4..d8700b1 100755
--- a/scripts/helper/create-topics.sh
+++ b/scripts/helper/create-topics.sh
@@ -10,6 +10,8 @@ echo "KAFKA_CLUSTER_ID: ${KAFKA_CLUSTER_ID}"

 auth="superUser:superUser"

+create_topic kafka1:8091 ${KAFKA_CLUSTER_ID} wikipedia.parsed.count-by-domain false ${auth}
+
 create_topic kafka1:8091 ${KAFKA_CLUSTER_ID} users true ${auth}
 create_topic kafka1:8091 ${KAFKA_CLUSTER_ID} wikipedia.parsed true ${auth}
 create_topic kafka1:8091 ${KAFKA_CLUSTER_ID} wikipedia.parsed.count-by-domain false ${auth}
```

... the script stops with:

```
ERROR: create topic failed {"error_code":40002,"message":"Topic 'wikipedia.parsed.count-by-domain' already exists."}
```

Is anything further required for fail-fast?

ybyzek: @javabrett this test does not fail fast:

```
[11:53:59] ~/git/cp-demo(topic-creation-via-rest-v3) ✗: git diff
diff --git a/scripts/helper/create-topics.sh b/scripts/helper/create-topics.sh
index 7adfca4..64f386a 100755
--- a/scripts/helper/create-topics.sh
+++ b/scripts/helper/create-topics.sh
@@ -8,7 +8,7 @@ source ${DIR}/functions.sh
 KAFKA_CLUSTER_ID=$(get_kafka_cluster_id_from_container)
 echo "KAFKA_CLUSTER_ID: ${KAFKA_CLUSTER_ID}"
 
-auth="superUser:superUser"
+auth="superUser:superUser2"
 
 create_topic kafka1:8091 ${KAFKA_CLUSTER_ID} users true ${auth}
 create_topic kafka1:8091 ${KAFKA_CLUSTER_ID} wikipedia.parsed true ${auth}
```

`stdout` shows:

```
{
  "servlet": "default",
  "message": "Unauthorized",
  "url": "/kafka/v3/clusters/AO6RqB0HSnaagGflWwIy4w/topics",
  "status": "401"
}
```

But the start script continues and does not exit

javabrett: @ybyzek OK I see what happens there - for auth 401 errors, the error response is actually different, so we don't detect the error.

`curl` is well-known in terms of difficulty capturing all of good responses, error responses and HTTP codes - do we have an idiomatic way of doing that?

ybyzek: I haven't played with this too much, but some suggestions to consider:

1. Parse output for "Unauthorized"
2. Check existence of topics afterwards

javabrett: @ybyzek I've pushed a commit which adds fetching of the HTTP response code from the `curl` command for `create_topic`, so we can test that in addition to the output scraping.  Unfortunately `curl` makes it fairly hard to fetch both output and response code, so the new bash script fragment is necessarily complicated.

The `create_topic` call now fails-fast for at least topic-already-exists (400) and unauthorized/bad-credentials (401).  Any response code >299 will be treated as fail.

ybyzek: @javabrett beauty:

```
ERROR: create topic failed {
"servlet":"default",
"message":"Unauthorized",
"url":"/kafka/v3/clusters/QKMFRq6RSbeOfk4-6G7pPw/topics",
"status":"401"
}
```

It fails fast!!

ybyzek (approved):
@javabrett LGTM -- wonderful improvement in time and showcases REST Proxy for topic creation -- win/win.  Thank you!

PR #307, opened 2020-12-09
DEVX-2320: Allow users to opt-out of Elasticsearch and Kibana to reduce memory footprint

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2320

Elasticsearch and Kibana are now opt-out:

* Allow option to reduce memory footprint of cp-demo for users who are memory constrained

Reasons to choose opt-out over opt-in:

* Viz is compelling to some users
* Default behavior (VIZ=true) is consistent with current behavior: continuity for existing users and no impact to jmx-monitoring-stacks (@jeanlouisboudart @waliaabhishek )
* Keeps Elasticsearch sink connector in main workflow: demonstrates & validates a sink connector
* Looking at Kibana dashboard simplifies smoke test to validate cp-demo is working after changes are made
* Smoother UX for future enhancements to visualizations added to cp-demo (cc: @robcowart )
* Simplifies docs

### Author Validation

- [x] Documentation: see https://github.com/confluentinc/docs-platform/pull/170 for staging server
- [x] Run cp-demo: both with default (`VIZ=true`) and `VIZ=false` 

### Reviewer Tasks

- [ ] Documentation: see https://github.com/confluentinc/docs-platform/pull/170 for staging server
- [ ] Run cp-demo: both with default (`VIZ=false`) and `VIZ=true` 
- [ ] jmx-monitoring-stacks: @jeanlouisboudart @waliaabhishek


jwfbean:
I'm imagining a world where we have swappable front ends. Can we call the variable something else, eg ES or ELASTIC?

ybyzek:
> I'm imagining a world where we have swappable front ends. Can we call the variable something else, eg ES or ELASTIC?

@jwfbean  While it's not a far-fetched idea, there are no plans for to swap another frontend.  Even more so, if there were a thought to future-proof, one could imagine using Elasticsearch but a different Viz tool (e.g. Grafana instead of Kibana).  It doesn't make sense to enumerate those permutations at this time, so my leaning is to leave it generic as `VIZ` for now.

jwfbean:
Does a cp-demo user ever want to use Elastic without Kibana? If yes, then `VIZ` is definitely a misnomer. If no, then yeah I agree it's fine.

ybyzek:
> Does a cp-demo user ever want to use Elastic without Kibana? If yes, then `VIZ` is definitely a misnomer. If no, then yeah I agree it's fine.

@jwfbean to the best of my knowledge, I haven't heard of this use case.

javabrett (approved):

mikebin (approved):
LGTM

jeanlouisboudart (approved):
LGTM

PR #310, opened 2020-12-14
Adds CORS properties to ksqlDB allowing access via Control Center

javabrett:
### Description 

Adds CORS properties required in ksqlDB Server configuration to allow it to be accessed via Control Center on a different server.

Prior to this change, accessing ksqlDB page in Control Center for CP 6.1.x branch will result in 401 UNAUTHORIZED for requests to `/ksql`.  With these properties set, the `wikipedia` application is no longer deactivated in the UI and can be clicked/accessed.

### Author Validation

- [ ] Documentation
- [X] Run cp-demo
- [ ] jmx-monitoring-stacks

### Reviewer Tasks

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -777,6 +777,12 @@ services:
       KSQL_BOOTSTRAP_SERVERS: kafka1:10091,kafka2:10092
       KSQL_HOST_NAME: ksqldb-server
       KSQL_LISTENERS: "http://0.0.0.0:8088,https://0.0.0.0:8089"
+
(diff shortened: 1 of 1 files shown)


ybyzek (changes requested):
@javabrett this PR is 🥇 But I'm marking as "Request changes" to intentionally block for a bit.  The reason is it's not clear from internal discussion (e.g. https://confluentinc.atlassian.net/browse/MMA-9429) whether a fix is going into the product that will do this by default, or whether PRs like this will be necessary for all ksqlDB properties.

PR #312, opened 2020-12-15
DEVX-2345: Add post-start checks

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2345

- Rudimentary postflight checks at the end of start.sh
- Intentional to not add wait time into the start.sh script
- Intentional to not `exit 1`

### Author Validation

- [x] Run cp-demo

### Reviewer Tasks

- [ ] Run cp-demo


rspurgeon:
@ybyzek the following feedback is based on indication this change is meant to assist with future testing automation, as captured in the linked issue under DEVX-2345.

"post-flight", to me, means something final where in this case the services are still running and the user is interacting w/ the demo.  What do you think about calling these particular checks "post-start" or "running" validation?  

In support of that suggestion, I could see wanting to later add different "post-flight" validations during or after the stop routine, to validate proper cleanup as we build some validation automations.

For automated validations, it might be better to produce error code(s) from the validation scripts so that they can be asserted against by the automation routines.  If you returned an error code from the "checks" script, it could be swallowed by the start.sh script, but also it could be useful to print an additonal warning to the demo user that the check script has produced an error independent of the individual warnings that script prints.

scripts/helper/functions.sh
```diff
@@ -59,6 +59,38 @@ preflight_checks()
...
+    echo -e "\nWARNING: at least one Docker container unexpectedly exited. Please troubleshoot, see https://docs.confluent.io/current/tutorials/cp-demo/docs/index.html#troubleshooting"
+  fi
+
+  # Validate connectors are running
+  # The Replicator connector may come fully UP after start script ends, so ignore that connector here (to not add time to start script)
+  if [[ "$VIZ" == "true" ]]; then
+    connectorList="wikipedia-sse elasticsearch-ksqldb"
+  else
+    connectorList="wikipedia-sse"
+  fi
```
rspurgeon, line 75: Could this be made a little more future proof by querying for the connector list from the connect worker and then querying for each status based on that result?

mikebin (commented):
Agree with @rspurgeon's feedback, otherwise LGTM.

scripts/helper/functions.sh
ybyzek: What if the connector was never successfully submitted, the connect worker wouldn't know about it.  The thinking is that the validation script should be directive, not let it to chance that the connector(s) were submitted properly.

ybyzek:
@rspurgeon thanks for the suggestions

1. Last commit renames `postflight` to `poststart`
2. wrt error codes, cp-demo can definitely do better.  FYI DEVX-2207 (automated testing for cp-demo) is a consideration but not the main driver for DEVX-2345.  The main driver of this PR is mostly to prevent some errors that I've seen come in from new users who don't even notice that some containers have stopped.  tl;dr my leaning is to not implement different error handling in this PR, but rather let it be handled by DEVX-2207 so it can be designed/implemented the right way.  WDYT?

scripts/helper/functions.sh
rspurgeon: I thought of this as well, but if the connector is missing in this case, are you going to get a 404 from the endpoint and will the STATE != "RUNNING" show something meaningful?  

Additionally, i'm thinking maybe it's too late for this error, and that situation should be highlighted by the submit script?

rspurgeon:
>     2. wrt error codes, cp-demo can definitely do better.  FYI DEVX-2207 (automated testing for cp-demo) is a consideration but not the main driver for DEVX-2345.  The main driver of this PR is mostly to prevent some errors that I've seen come in from new users who don't even notice that some containers have stopped.  tl;dr my leaning is to not implement different error handling in this PR, but rather let it be handled by DEVX-2207 so it can be designed/implemented the right way.  WDYT?

Pushing more robust error handling to DEVX-2207 works for me

scripts/helper/functions.sh
rspurgeon: We don't have to hold up this PR for this feedback, but we might want to double check what happens if the connector is not there (test w/ an invalid name maybe?).

ybyzek: Refactored code to check status of connectors configured in connect

rspurgeon (approved):
LGTM

PR #323, opened 2021-01-11
DEVX-2256: Remove and replace deprecated configurations

rspurgeon:
### Description 
Removes or replaces deprecated configuraitons for SR, connect, and Kafka REST proxy.

In the case of key converters, the unecessary configurations are removed.  in the case of SSL client auth, the proper configuration is put in place of the deprecated name.

### Author Validation

- [X] Run cp-demo

### Reviewer Tasks

- [ ] Run cp-demo


awalther28 (approved):
I validated this PR by running cp-demo. I checked that the schema registry config `ssl.client.authentication` was being set as expected to `NONE`.

mikebin (approved):
LGTM. Would it make sense to set the brokers' `ssl.client.auth=requested`, since SSL authN is only needed for the SSL listener, but not for the TOKEN listener?

ybyzek (approved):
@rspurgeon LGTM, thanks!

rspurgeon:
> LGTM. Would it make sense to set the brokers' `ssl.client.auth=requested`, since SSL authN is only needed for the SSL listener, but not for the TOKEN listener?

@mikebin The latest has this change, can you please validate this is proper?  I'm not familiar with this area of the product.  Thank you

ybyzek:
@rspurgeon I ran cp-demo with the latest and it works.

@mikebin for my education, if `SSL authN is only needed for the SSL listener, but not for the TOKEN listener`, then how did `required` work?  

Is it that:

1. `SSL authN` codepath was never invoked for the `TOKEN listener` anyway, or
2. `SSL authN` codepath was invoked for the `TOKEN listener` but it was redundant, or
3. I am missing another piece of the puzzle

mikebin:
> Is it that:
SSL authN codepath was never invoked for the TOKEN listener anyway, or
SSL authN codepath was invoked for the TOKEN listener but it was redundant, or
I am missing another piece of the puzzle

PR #335, opened 2021-01-22
ST-4894: Adding Jenkinsfile

elismaga:
### Description 

### Author Validation

### Reviewer Tasks

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,11 +20,14 @@ def job = {
                 LOGIN_CMD=$(aws ecr get-login --no-include-email --region us-west-2)
                 $LOGIN_CMD
             '''
-
-            sh '''
-                export REPOSITORY=368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc
-                ./scripts/start.sh
-            '''
+            try {
+                sh '''
+                    export REPOSITORY=368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc
+                    ./scripts/start.sh
+                '''
+            } finally {
+                sh '''docker-compose logs zookeeper'''
+            }
         }
     }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,6 +27,8 @@ def job = {
                 '''
             } finally {
                 sh '''docker-compose logs zookeeper'''
+                sh '''netstat -ltnp | grep -w ":2181"'''
+                sh '''lsof -i :2181''' 
             }
         }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ def config = jobConfig(defaultConfig)
 
 def job = {
     stage("Startup") {
-//        withDockerServer([uri: dockerHost()]) {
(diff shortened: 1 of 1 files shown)


ybyzek:
@elismaga can this PR target this for `6.1.x` instead of `master` so that it can work on the upcoming release of CP 6.1.0?

Jenkinsfile
```diff
@@ -0,0 +1,45 @@
...
+
+            sh '''
+                export REPOSITORY=368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc
+                export CONFLUENT=6.1.0-0
+                export CONFLUENT_DOCKER_TAG=6.1.x-latest
+                export CONFLUENT_SHORT=6.1
+                export CONFLUENT_MAJOR=6
+                export CONFLUENT_MINOR=1
+                export CONFLUENT_PATCH=0
+                ./scripts/start.sh
```
rspurgeon, line 32: An FYI: `./scripts/start.sh` sources in the config.env file which, as you know, is managed by the tools-eng automation.  

Ideally we are relying on that as much as possible, so this works across the branches.  I think we'll need a way to set the REPOSTIORY variable optionally on execution, depending on the branch.  For .x branches we want to be able to pull from an internal repo and otherwise public docker hub.

Also, just dropping a thought here to be sure... are we ok with pushing to this public GH repo the Confluent internal ECR account and address?  Or instead can we just treat it like other secrets?  It has been circulated, mostly in the examples repo, in the past I'm just not sure it would be preferred in here.

env_files/config.env
```diff
@@ -5,13 +5,13 @@
...
-CONFLUENT_DOCKER_TAG=6.2.x-latest
-CONFLUENT_SHORT=6.2
+CONFLUENT=6.1.0-0
+CONFLUENT_DOCKER_TAG=6.1.x-latest
+CONFLUENT_SHORT=6.1
 CONFLUENT_PREVIOUS=""
 CONFLUENT_RELEASE_TAG_OR_BRANCH=master
 CONFLUENT_MAJOR=6
-CONFLUENT_MINOR=2
+CONFLUENT_MINOR=1
```
rspurgeon, line 14: Sorry if i'm jumping the gun, I know this marked Draft, but i don't think any change should modify this file

elismaga:
> @elismaga can this PR target this for `6.1.x` instead of `master` so that it can work on the upcoming release of CP 6.1.0?

@ybyzek yeah, i am planning on doing that today

env_files/config.env
elismaga: @rspurgeon i wasn't planning on merging this change, it was just for testing on this branch.

Jenkinsfile
elismaga: @rspurgeon having that repository variable only get set based on the branch being run isn't an issue. i don't think having the ECR address be in a public file is an issue, as i think other public scripts already have this in them. I will double check though. The credentials for the account are not exposed by doing this.

chuck-confluent:
I am going to close this for now.

PR #339, opened 2021-02-02
DEVX-2491: Add explicit Docker pull script to cp-demo and add to docu…

ybyzek:
…mentation that it takes a long time

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2491

We can't change download time but:

- Add to documentation to warn first-time users that it will take a long time, to set user expectations
- Add script to just download Docker images
- Add to documentation note of this script

### Author Validation

- [x] Run cp-demo

### Reviewer Tasks

- [ ] Run cp-demo


awalther28 (commented):
I ran cp-demo by first running the docker-images-pull.sh, then running start.sh. It works as expected. @ybyzek one thing that worries me about this is that now we have 2 places to update the the non-cp image versions (ie two places to update elasticsearch version). In an ideal world, I would think we'd want those versions residing in a variable to make maintenance less error prone. Thoughts?
Generally, why introduce a docker-pull script? After reading DEVX-2491, is the idea to better level set that cp-demo will download a number of things?

ybyzek:
@awalther28 yes, the idea was about level setting expectations that pulling down 10+GB of Docker images will take a long time.  

I understand the concern about maintaining versions in multiple places.  I have no qualms about nuking the script, and just making this PR only about adding a note into the docs about long download times for Docker images.  WDYT?

awalther28:
@ybyzek If you want to keep the script, is there an easy way to set the container versions in a singular place? Maybe config.env--you know this project structure much better than I.

I think changing the docs could suffice, like changing this line ("1. To run cp-demo the first time with defaults, run the following command. This takes a few minutes to complete.") to something more explicit like "1. ...This will take at minimum 10 minutes and will download 13 container images". 

Again, you know this project like the back of your hand so I differ to you on whether or not to keep this script or just change the docs.

ybyzek:
@awalther28 let's keep it to docs only... please see trimmed down PR in latest commit

awalther28 (approved):
Thanks @ybyzek LGTM

PR #345, opened 2021-02-22
DEVX-2482: Clarify ZooKeeper TLS 1.3 support

javabrett:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2482

- Removes separate config and associated comment for `ZOOKEEPER_SSL_CIPHER_SUITES`.  ZooKeeper can use the same set of ciphersuites as other components, but the TLS 1.3 suites will be automatically filtered-out by default.
- Added commented, commented-out configuration showing that Broker->ZooKeeper will use TLS 1.2 by-default, which is supported and tested, but with instructions on how to configure TLS 1.3 if desired - this should be considered experimental and self-tested.  TLS 1.3 support in ZooKeeper is very new and there may be issues.

### Author Validation

[ ] Documentation
[X] Run cp-demo
[ ] jmx-monitoring-stacks

### Reviewer Tasks


ybyzek (approved):
@javabrett validated cp-demo run, LGTM

PR #347, opened 2021-02-25
Update docker-compose.yml

mcascallares:
Exporting UDF and pull queries metrics

### Description 

Exposing new metrics for pull queries and UDFs

### Author Validation

### Reviewer Tasks


ybyzek:
Thanks for the PR @mcascallares .  

1. What is the resource impact of this PR?  cp-demo is already a heavy-weight so we have to take into consideration any additional resources that might be required.
2. How can this be used by end-users?  How are the metrics exposed?

mcascallares:
My pleasure @ybyzek. Thank you for your promptly reply!

1. New metrics are exposed by ksqldb-server via JMX. There is no direct impact on the resource utilization.
2. It would be exposed by external consumers, for example by https://github.com/confluentinc/jmx-monitoring-stacks. I started to explore some of the pull queries metrics using that project in the Prometheus + Grafana flavour (WIP).

mcascallares:
Hi again @ybyze!

I was checking with @jeanlouisboudart, and another approach would be to override that variable in https://github.com/confluentinc/jmx-monitoring-stacks/blob/6.0.0-post/jolokia-elastic-kibana/docker-compose.override.yml and do not change that setting in cp-demo. Up to you.

Thanks

ybyzek:
If the way that this is utilized is primarily via `jmx-monitoring-stacks`, then I would agree with the approach by @jeanlouisboudart -- this would be better suited as a PR to that repo.

mcascallares:
Sounds good, I will close this PR.
Thanks!

PR #348, opened 2021-03-02
DEVX-2533: training-independent tools image; DEVX-2534: optimize logi…

ybyzek:
…n by adding CA cert

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2533
https://confluentinc.atlassian.net/browse/DEVX-2534

- Use training-independent tools image
- Add CA cert to root in tools image to optimize CLI login

### Author Validation

- [x] Run cp-demo: including ccloud portion to validate `date`

### Reviewer Tasks

### DIFF
--- a/tools/Dockerfile-tools
+++ b/tools/Dockerfile-tools
@@ -1,0 +1,19 @@
+FROM ubuntu:18.04
+
+ENV CONFLUENT_HOME=/usr \
+    TERM=xterm-256color
+
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        git \
+        curl \
+        wget \
+        jq \
+        netcat \
+        httpie \
+        peco \
+        vim \
+        expect \
+    && curl --http1.1 -L https://cnfl.io/cli | sh -s -- -b /usr/bin \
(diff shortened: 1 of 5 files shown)


huynconfluent (approved):
LGTM

rspurgeon (approved):
LGTM

chuck-confluent (approved):
It looks good to me, but I can't see why the CI build failed in semaphor (I don't think I have permissions)

tools/Dockerfile-tools
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:18.04
```
javabrett, line 1: Noting `ubuntu` base image, which is certainly familiar, but other images have moved to UBI-8.  Is there any value is using same here?

ybyzek: @javabrett the reason this PR currently uses `ubuntu` is because the original `training-tools` Docker image had the same, and I wanted to ensure **exactly** the same behavior with packages like `date`.  I don't think there is a strong reason to have consistency with the CP images, and I don't have time to explore other options, so I prefer to leave this PR as-is with `ubuntu` unless there is a downside I'm not aware of.

javabrett (approved):

PR #350, opened 2021-03-08
Set Schema and Secret registries (_schemas, _confluent-secrets) topics RF=2

javabrett:
### Description 

We already set `...TOPIC_REPLICATION_FACTOR=2` for many auto-created/internal topics, and also for the user/application topics created.  This provides a consistent replication factor for all topics of 2, which also equals the normal broker count.

This is also useful if you want to modify cp-demo and add another broker (2->3) before start-up, but still only want RF=2 for all topics.  Certain auto or internal topics will default to RF=3 if three brokers are available during startup when they are created.  This is OK, but it does block later removal of that third broker if any topics have RF=3.

This PR sets RF=2 for internal topics `_schemas` and `_confluent-secrets` so that they are consistent with all other topics and if >2 brokers are there for the initial start-up, brokers can be removed.

### Author Validation

- [ ] Documentation
- [X] Run cp-demo
- [ ] jmx-monitoring-stacks

This can be a little tricky to test - let me know if you want a diff which will launch a third broker at first startup ... that can then show the RF=3 before this patch and RF=2 after, on both those topics.

### Reviewer Tasks


ybyzek:
@javabrett the following test for this PR yields strange results.  I pulled this branch and changed RF to 4 for those 2 lines:

```
[8:01:58] ~/git/cp-demo(schemas-and-secrets-topic-replication-factor-2) ✗: git diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 6a442ba..c33ed1b 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -561,7 +561,7 @@ services:
       CONNECT_CONFIG_PROVIDERS_SECRET_CLASS: 'io.confluent.connect.secretregistry.rbac.config.provider.InternalSecretConfigProvider'
       CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_MASTER_ENCRYPTION_KEY: 'password1234'
       CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_KAFKASTORE_BOOTSTRAP_SERVERS: kafka1:10091,kafka2:10092
-      CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_KAFKASTORE_TOPIC_REPLICATION_FACTOR: 2
+      CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_KAFKASTORE_TOPIC_REPLICATION_FACTOR: 4
       CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_KAFKASTORE_SECURITY_PROTOCOL: SASL_SSL
       CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_KAFKASTORE_SSL_TRUSTSTORE_LOCATION: /etc/kafka/secrets/kafka.connect.truststore.jks
       CONNECT_CONFIG_PROVIDERS_SECRET_PARAM_KAFKASTORE_SSL_TRUSTSTORE_PASSWORD: confluent
@@ -747,7 +747,7 @@ services:
       SCHEMA_REGISTRY_SCHEMA_REGISTRY_INTER_INSTANCE_PROTOCOL: "https"
       SCHEMA_REGISTRY_LOG4J_ROOT_LOGLEVEL: INFO
       SCHEMA_REGISTRY_KAFKASTORE_TOPIC: _schemas
-      SCHEMA_REGISTRY_KAFKASTORE_TOPIC_REPLICATION_FACTOR: 2
+      SCHEMA_REGISTRY_KAFKASTORE_TOPIC_REPLICATION_FACTOR: 4
       SCHEMA_REGISTRY_SSL_CIPHER_SUITES: ${SSL_CIPHER_SUITES}
 
       SCHEMA_REGISTRY_DEBUG: 'true'
```

Then ran cp-demo with 2 brokers (as it is today)...no 3rd broker.

Expectation was that this would fail, but it does not.  For example, SR logs show that the config is recognized, RF is set to 4

```
schemaregistry    | 	kafkastore.topic.replication.factor = 4
```

However, the` _schemas` topic still has 2 replicas:

```
[8:03:59] ~/git/cp-demo(schemas-and-secrets-topic-replication-factor-2) ✗: docker-compose exec kafka1 kafka-topics \             
              --describe --topic _schemas \
              --bootstrap-server kafka1:12091
Topic: _schemas	PartitionCount: 1	ReplicationFactor: 2	Configs: cleanup.policy=compact
	Topic: _schemas	Partition: 0	Leader: 2	Replicas: 2,1	Isr: 2,1	Offline: 
```

Either SR itself is resetting RF from 4 to 2 because of broker availability, or this config is being overwritten by some other config?  Regardless, would be good to understand so that we know that this PR is effective in what it intends.

ybyzek (approved):
@javabrett LGTM

javabrett:
Yep I just ran the same RF=4 test with the same logged messages and results ... both topics with a log message and end up with RF=2 (== num brokers).  The main aim here is if you start with 3 brokers, you still retain RF=2 on these topics, allowing a broker to be removed without causing permanent URPs.

Thanks for the review, merging.

PR #353, opened 2021-03-12
DEVX-2162: add demo artifacts for Self Balancing Clusters (SBC)

javabrett:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2162

- Adds configuration and scripting for a third broker, `kafka3` which can join the cluster but is not started by-default.
- The partition count for topics is increased from 2 to 6 to better demonstrate cluster-rebalancing.  
- Self Balancing Cluster configuration for self-healing on broker failure is reduced (for demo) from 1 hour to 30 seconds.
- Scripts are added to start broker `kafka3` and monitor the rebalancing logs, and to kill `kafka3` and monitor the logs while cluster healing is underway.

### Author Validation

- [ ] Documentation
- [X] Run cp-demo
- [ ] jmx-monitoring-stacks

### How to test

- Start `cp-demo` as normal.
- Once startup has completed, you can either proceed with the SBC add-broker demo immediately, or wait around 15 minutes.  If you start immediately, an add-broker task will register immediately and be visible, but there will be a 15 minute or so delay while SBC collects statistics for the rebalance plan.
- Run `scripts/sbc/add-broker.sh` which will start broker `kafka3` then wait-with-timeout for SBC to log a message that it has created a rebalance task.
- Visit the Control Center UI and review the Self Balancing Cluster summary panel and see an Add Broker task appear.
- Run `docker-compose exec kafka1 kafka-topics   --describe   --topic pageviews --bootstrap-server kafka1:9091   --command-config /etc/kafka/secrets/client_sasl_plain.config` - use `watch` if desired ... look for eventual partitions, leaders and replicas being moved onto broker 3.
- Note Add Broker task completes - check with scripts `scripts/validate/validate_sbc_add_broker_reassignment.sh` and `scripts/validate/validate_sbc_add_broker_completed.sh`.

Test broker removal - option 1, scheduled removal, scale-down

- Visit broker 3 in the Control Center UI and use the Remove Broker option.
- Tail logs: `docker-compose logs -f --tail=1000 | grep -E "(databalancer|cruisecontrol)" | grep -v "confluent.balancer.class"`
- Run `docker-compose exec kafka1 kafka-topics   --describe   --topic pageviews --bootstrap-server kafka1:9091   --command-config /etc/kafka/secrets/client_sasl_plain.config` and note under-replicated partitions.  See these also in the Control Center UI.
- SBC will eventually complete and rebalance the partitions that were on broker 3 across brokers 1 and 2.
- Run `kafka-topics` again and check no URPs remain.

Test broker removal - option 2, unscheduled

- Run `scripts/sbc/kill-broker.sh`, which will stop broker `kafka3` and then watch for SBC to log a message that self-healing has commenced.
- SBC will take 30 seconds to act on lost broker.
- Perform similar monitoring to scheduled removal - URPs should be 0 when rebalancing has completed.
- Validate completion some time later with `scripts/validate/validate_sbc_kill_broker_completed.sh`.

### Reviewer Tasks


scripts/sbc-add-broker.sh
```diff
@@ -0,0 +1,14 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+source ${DIR}/helper/functions.sh
+source ${DIR}/env.sh
+
+#-------------------------------------------------------------------------------
+
+echo "Starting broker kafka3, then tailing SBC logging output.  Type CTRL-C to stop tailing logs."
+docker-compose up -d kafka3
+docker-compose logs -f --tail=1000 | grep -E "(databalancer|cruisecontrol)" | grep -v "confluent.balancer.class"
```
ybyzek, line 14: I'm inclined to not have a script that just keeps running with a log tail.  For consideration: maybe it's better to do something like (pseudo code):

1. bring up kafka3
2. set a wait time and run `retry` to look for a specific message in the log
3. exit when log message is seen or max wait time is reached

scripts/sbc-kill-broker.sh
```diff
@@ -0,0 +1,14 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+source ${DIR}/helper/functions.sh
+source ${DIR}/env.sh
+
+#-------------------------------------------------------------------------------
+
+echo "Killing broker kafka3, then tailing SBC logging output.  SBC will commence healing after about 30 seconds.  Type CTRL-C to stop tailing logs."
+docker stop kafka3
+docker-compose logs -f --tail=1000 | grep -E "(databalancer|cruisecontrol)" | grep -v "confluent.balancer.class"
```
ybyzek, line 14: Same comment as above wrt script running with a log tail

scripts/sbc-kill-broker.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
ybyzek, line 1: Not sure what the total sum of SBC related scripts will be, but WDYT about creating a distinct folder `scripts/sbc/` and have these scripts live there?

ybyzek (commented):
Thanks for the PR @javabrett .  Left a few comments.  Also, is it the intention to create a separate PR for docs (and were you thinking I would do that, or you plan to?)?

docker-compose.yml
```diff
@@ -418,6 +420,178 @@ services:
       KAFKA_KAFKA_REST_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS: 60000
       KAFKA_KAFKA_REST_CLIENT_CONFLUENT_METADATA_SERVER_URLS_MAX_AGE_MS: 60000
         
+  kafka3:
```
ybyzek, line 423: Hmmm, my comment got lost, I'll add again.  I think this PR should enable SBC with minimal impact to current demo, so I propose making the following changes:

- create a new folder `scripts/sbc`
- move add/kill script there
- use `certs-create-one-user.sh` instead of modifying `certs-create.sh`
- use a separate docker-compose file for kafka3 instead of modifying this base one

And it really brings into question whether the partition count should be modified in this PR or not, because that changes _does_ modify the base demo and everything needs to be re-validated end-to-end (code, docs, screenshots)

scripts/helper/topic.jq
```diff
@@ -1,6 +1,6 @@
 {
     "topic_name": $topic_name,
-    "partitions_count": 2,
+    "partitions_count": 6,
```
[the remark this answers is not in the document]
javabrett: I agree with this and comments further on that this change isn't essential for this demo - it does mean that you can't say that each topic-partition now makes full use of the third broker _individually_, but the cluster is still rebalanced globally.  We can bump this change until a later/larger testing cycle.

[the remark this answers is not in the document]
javabrett: Pushed a commit to revert this change.

javabrett:
> Thanks for the PR @javabrett . Left a few comments. Also, is it the intention to create a separate PR for docs (and were you thinking I would do that, or you plan to?)?

I'm happy to write/doc the demo path - either as a following PR (if this PR ends up looking like something that can be merged pending docs), or as following commits on this PR.  Just want to get the structure agreed first.

scripts/sbc-add-broker.sh
javabrett: Agreed, the foreground log tail isn't great.

The main motivation for the script in the first place, is that we don't have first-class support for demo lifecycle outside of `start.sh` and `stop.sh`, so the script sources `env.sh` to avoid needing to do that manually.  Then there's some monitoring support in Control Center but the logs provide a lot of details.  The SBC process can be long-running, so I'm not sure it fits log-monitor retry/timeout perfectly ... feels more like something that the curious can monitor, with perhaps some better log filtering.

I'm somewhat inclined to drop the script altogether and do this via documented command, like:

- `(source scripts/env.sh;docker-compose up -d kafka3)`
- `docker-compose logs -f | grep "com.linkedin.kafka.cruisecontrol.executor.Executor"`

... which can be used to optionally follow SBC progress.

The documentation will describe the required `kafka-topics` commands to run to see the changes made by SBC once Control Center and logs confirm completion.

WDYT?

scripts/sbc-kill-broker.sh
javabrett: Agreed - if we retain the scripts I will move them to `scripts/sbc` ... now considering though whether we need them at all, or just some documented demo steps/commands.

ybyzek: @javabrett IMO the scripts are helpful to have for validation.  For example, pre-GA as the CP releases are rev'd, I use [validation scripts](https://github.com/confluentinc/cp-demo/tree/6.1.0-post/scripts/validate) as a sanity test to verify everything is working.  It's a poor man's version of what a full testing workflow should be, but it's pretty effective to make sure the more complicated pieces of cp-demo work...beyond just a successful `scripts/start.sh` run.

So maybe we can think about docs decoupled from validation: what you decide to put into docs for SBC may look different than what you decide to put into a validation script for SBC.  You could omit the script, but I'd love to have _something_ if possible as a quick sanity check.

scripts/sbc-add-broker.sh
ybyzek: Please see https://github.com/confluentinc/cp-demo/pull/353#discussion_r594280310 for comment on above.

docker-compose.yml
javabrett: Any room to negotiate on the don't modify `certs-create.sh` suggestion?

This one has the most potential to create problems in the workflow, since I assume this means that we would generate the cert for `kafka3` on-demand, or late, just prior to starting the `kafka3` broker as part of an SBC test-run.  It looks simpler to generate all the certs at the same time and build the keystores once.  Adding `kafka3` here amounts to creating a spare cert for the broker to use later ... if it's a significant concern I can look into how to safely extract this into its own lifecycle.

javabrett:
Rebased to current 6.1.0-post and added new commits, summary:
- Reverted change 2 partitions -> 6 partitions ... not a necessary change for this demo.
- Added validate scripts for add-broker (planning, reassigning, completed) and kill-broker (started, completed).  The started validations are called by the add/kill broker scripts, others should be called at various times on-demand.
- Moved SBC test scripts to `scripts/sbc/`.
- Moved `kafka3` into its own Docker Compose file.

Updated PR test-plan.

Have not yet stopped `kafka3` certificate pre-generation along with other services.

docker-compose.yml
ybyzek: The reasoning to create the `kafka3` cert later was to clearly articulate the boundary between the cp-demo base from the add-on SBC module, to not confuse folks who aren't exercising the SBC workflow, and to avoid unexpected problems down the road.  Since there is already a bash script to bring up the third broker, it would be a one line addition into that script to create the cert.  It's not a significant concern, NOT making the change won't block the PR, but I'm curious for the reasoning to not do this?

javabrett: Pushed a commit to separate `kafka3` certificate-generation.  Initial change for this had a bug which I've now fixed.

scripts/sbc/README.md
```diff
@@ -0,0 +1,3 @@
+# cp-demo Self Balancing Clusters
+
```
ybyzek, line 2: @javabrett I understand that at some point docs may follow, but until then, perhaps the README can mention the scripts to run?  e.g. very brief mention of the key scripts, 

```
Run these scripts:

`./scripts/sbc/add-broker.sh`: add 3rd broker to force rebalancing
`./scripts/sbc/kill-broker.sh`: remove 3rd broker
```

WDYT?

scripts/sbc/add-broker.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+IFS=$'\n\t'
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+source ${DIR}/../helper/functions.sh
+source ${DIR}/../env.sh
+
+#-------------------------------------------------------------------------------
+
+(cd $DIR/security && ./certs-create-per-user.sh kafka3) || exit 1
```
ybyzek, line 12: Thank you!

scripts/sbc/README.md
javabrett: Excellent idea, done.

ybyzek (approved):
@javabrett thanks for addressing all feedback, it's really clean now with all the changes isolated to the `scripts/sbc/` folder.  LGTM!

Update: I missed that the validation scripts are in the validation folder, not sbc, but I understand the reasoning for those files being in that folder.  Still LGTM

ybyzek (changes requested):
I'm thinking about this more -- I think the validation scripts aren't actually in the right location.

scripts/validation -- these are whole scripts that are the "smoke tests" that can be run after cp-demo script completes, they are not meant to checkpoint individuals stages of a workflow as done here.

I'm really sorry for changing status from `approved` to `request changes` :( but I think this change should be made to ensure clear delineation and intention of scripts.  I'll submit a PR to your branch...if that's ok

ybyzek (approved):
Beauty, and approved!  Thanks @javabrett 👍

PR #369, opened 2021-06-03
Adding placeholders for Telemetry Reporter

jessemillerjom:
Adding in placeholder config setting on kafka1 and kafka2 for the Confluent Telemetry Reporter to make it easier for users to get started with Confluent Health+

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2620

This PR provides commented out config parameters for the Confluent Telemetry Reporter. 

### Author Validation

### Reviewer Tasks


ybyzek:
@jessemillerjom this might make sense in cp-all-in-one (which is a launchpad for other work) but seems like a less fit for here.  cp-demo has its own workflow for Telemetry Reporter (see https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/hybrid-cloud.html#telemetry-reporter) so it should not be in the `docker-compose.yml`.

Issue #371, opened 2021-06-10
non-descriptive start.sh error message if docker daemon isn't running


blaisdellk:
**Description**
If the docker daemon is not running start.sh will fail with
`panic: reflect: indirection through nil pointer to embedded struct [recovered]
	panic: reflect: indirection through nil pointer to embedded struct
`

**Troubleshooting**
Check if the docker daemon is running.  If it is stop it.
Execute scripts/start.sh

**Environment**
 - GitHub branch: 6.2.0-post
 - Operating System: MacOS Catalina
 - Version of Docker: Docker Engine Community 20.10.5
 - Version of Docker Compose:  1.29.0, build 07737305

PR #372, opened 2021-06-10
verify docker daemon is running in preflight checks

blaisdellk:
### Description 

https://github.com/confluentinc/cp-demo/issues/371

Verify docker daemon is running.

### Author Validation

[ x] Run cp-demo (MacOS Catalina only)

### Reviewer Tasks

 test on linux

### DIFF
--- a/scripts/helper/functions.sh
+++ b/scripts/helper/functions.sh
@@ -41,6 +41,9 @@ preflight_checks()
     verify_installed $cmd || exit 1
   done
 
+  # Verify Docker daemon is running
+  docker ps -q || exit 1
+
   # Verify Docker memory is at least 8 GB
   if [[ $(docker system info --format '{{.MemTotal}}') -lt 8000000000 ]]; then
     echo -e "\nWARNING: Memory available to Docker should be at least 8 GB (default is 2 GB), otherwise cp-demo may not work properly.\n"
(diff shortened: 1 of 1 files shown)


scripts/helper/functions.sh
```diff
@@ -41,6 +41,9 @@ preflight_checks()
     verify_installed $cmd || exit 1
   done
 
+  # Verify Docker daemon is running
+  docker ps -q || exit 1
```
ybyzek, line 45: @blaisdellk if it's just an exit, user may not know why. Please add an error message (see above/below) for examples.  Perhaps something like `ERROR: cp-demo requires Docker to be running.  Please start Docker and run again.`

ybyzek: @blaisdellk makes sense.

ybyzek (commented):
@blaisdellk thanks for the PR, LGTM.  Before approving, I just wanted to highlight that the base branch being merged into is `master`, but most users are running latest (`6.2.0-post` at the moment).  As a result, users won't see this change until a future release.  Unless: consider applying this PR to `6.2.0-post` instead.

blaisdellk:
Thanks @ybyzek.  I'm not aware of a way to alter the GitHub PR target so I created a https://github.com/confluentinc/cp-demo/pull/374 to supersede this PR.

PR #373, opened 2021-06-16
Fix for DEVX-2639

nsharma-git:
### Description 
cp-demo looks for kafka-connect-replicator:6.0.3 on branch 6.0.x
Since this plugin is not available on confluent hub, the demo fails to start. 

Since 6.0.3 is a security patch, we can use the connectors from previous version

### Author Validation

[ ] Run cp-demo


ybyzek (changes requested):
@nsharma-git this results in differing versions of Replicator & Connect, which is not supported.

> Replicator and Connect must always be running on the same version, we don’t support different versions of Replicator and Connect. If it’s working now that may be a coincidence.

nsharma-git:
> @nsharma-git this results in differing versions of Replicator & Connect, which is not supported.
> 
> > Replicator and Connect must always be running on the same version, we don’t support different versions of Replicator and Connect. If it’s working now that may be a coincidence.

Sorry for my misunderstanding. From earlier configuration, I was under impression that its supported configurations - https://github.com/confluentinc/cp-demo/blob/5.5.x/env_files/config.env

Issue #367, opened 2021-04-15
MDS fails to start due to wrong Java version


ybyzek:
**Description**

Start script failure:

```
Waiting up to 120 seconds for MDS to start
........................ERROR: Failed after 120 seconds. Please troubleshoot and run again. 
```

Docker logs:

```
[2021-04-14 20:35:00,087] WARN Failed to initialize a channel. Closing: [id: 0x67e8d0a5] (io.netty.channel.ChannelInitializer)
kafka2      | org.apache.zookeeper.common.X509Exception$SSLContextException: Failed to create KeyManager
kafka2      | 	at org.apache.zookeeper.common.X509Util.createSSLContextAndOptions(X509Util.java:336)
kafka2      | 	at org.apache.zookeeper.common.X509Util.createSSLContext(X509Util.java:268)
kafka2      | 	at org.apache.zookeeper.ClientCnxnSocketNetty$ZKClientPipelineFactory.initSSL(ClientCnxnSocketNetty.java:454)
kafka2      | 	at org.apache.zookeeper.ClientCnxnSocketNetty$ZKClientPipelineFactory.initChannel(ClientCnxnSocketNetty.java:444)
kafka2      | 	at org.apache.zookeeper.ClientCnxnSocketNetty$ZKClientPipelineFactory.initChannel(ClientCnxnSocketNetty.java:429)
kafka2      | 	at io.netty.channel.ChannelInitializer.initChannel(ChannelInitializer.java:129)
kafka2      | 	at io.netty.channel.ChannelInitializer.handlerAdded(ChannelInitializer.java:112)
kafka2      | 	at io.netty.channel.AbstractChannelHandlerContext.callHandlerAdded(AbstractChannelHandlerContext.java:938)
kafka2      | 	at io.netty.channel.DefaultChannelPipeline.callHandlerAdded0(DefaultChannelPipeline.java:609)
kafka2      | 	at io.netty.channel.DefaultChannelPipeline.access$100(DefaultChannelPipeline.java:46)
kafka2      | 	at io.netty.channel.DefaultChannelPipeline$PendingHandlerAddedTask.execute(DefaultChannelPipeline.java:1463)
kafka2      | 	at io.netty.channel.DefaultChannelPipeline.callHandlerAddedForAllHandlers(DefaultChannelPipeline.java:1115)
kafka2      | 	at io.netty.channel.DefaultChannelPipeline.invokeHandlerAddedIfNeeded(DefaultChannelPipeline.java:650)
[...]

PR #382, opened 2021-08-02
Move security commands to Docker: #367

ybyzek:
### Description 

#367 

See issue for details, but the solution in this PR is to move the certificate creation commands to a Docker image, to remove dependency on Java version in host environment

Note, to validate this PR, it's essential to run the start script with `CLEAN=true`

### Author Validation

- [x] Run cp-demo: as part of the validation, I downloaded Java v16 and verified that current cp-demo fails, then ran cp-demo with this PR and it passes.

### Reviewer Tasks

- [x] Run cp-demo


rspurgeon:
@ybyzek FYC, instead of adding another docker image definition, and docker-compose service to maintain (and another image for the user to pull down), WDYT about using the existing kafka image that we already have?  Mount a script and the volumes to a `docker run` for `cp-server:$CONFLUENT_DOCKER_TAG`.  Added benefit is we will use whatever version of Java the CP images have instead of any potential issue of hardcoding the java version 8 in the new local docker image definition.

ybyzek:
> @ybyzek FYC, instead of adding another docker image definition, and docker-compose service to maintain (and another image for the user to pull down), WDYT about using the existing kafka image that we already have? Mount a script and the volumes to a `docker run` for `cp-server:$CONFLUENT_DOCKER_TAG`. Added benefit is we will use whatever version of Java the CP images have instead of any potential issue of hardcoding the java version 8 in the new local docker image definition.

@rspurgeon excellent suggestion!  I have implemented it in the latest version of this PR.

rspurgeon (approved):
LGTM 🥇 tested w/ a cp-demo run, all systems go

javabrett (approved):
OK with me.

Main downside is that whenever certs need (re)generation we have to pay for a `yum -y install findutils;` into a temp image - I guess that's worth it for consistency.  Not important for now, but I wonder if there's just some old regular image that already includes the right Java version and `find`.

ybyzek:
@javabrett the cost of `yum -y install findutils` for a workflow that doesn't happen that often seems ok to incur, but just to document in the PR if needed for later, one alternative could be to build a local image using cp-server as the base, adding yum install, then saving that image locally for reuse

PR #386, opened 2021-08-30
[ci-skip] Set CLASSPATH Monitoring Interceptor jar to wildcard to catch any version (5.4.x only)

imcdo:
### Description 
The class `MonitoringConsumerInterceptor` is not in the classpath by default, and is only available on the schema registry container.   This update adds the internceptor to the classpath so that it can be used by the consumer.

### Author Validation
ran cp-demo start consumer script, validated that the consumergroup starts up

- [x] Run cp-demo

### Reviewer Tasks

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -355,7 +355,7 @@ services:
       CONNECT_PLUGIN_PATH: "/usr/share/java,/connect-plugins"
       CONNECT_LOG4J_ROOT_LOGLEVEL: INFO
       CONNECT_LOG4J_LOGGERS: org.reflections=ERROR
-      CLASSPATH: /usr/share/java/monitoring-interceptors/monitoring-interceptors-*.jar
+      CLASSPATH: /usr/share/java/monitoring-interceptors/*
 
       # Connect Worker
       CONNECT_SECURITY_PROTOCOL: SASL_SSL
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -355,7 +355,7 @@ services:
       CONNECT_PLUGIN_PATH: "/usr/share/java,/connect-plugins"
       CONNECT_LOG4J_ROOT_LOGLEVEL: INFO
       CONNECT_LOG4J_LOGGERS: org.reflections=ERROR
-      CLASSPATH: /usr/share/java/monitoring-interceptors/monitoring-interceptors-5.4.5-SNAPSHOT.jar
+      CLASSPATH: /usr/share/java/monitoring-interceptors/monitoring-interceptors-*.jar
 
       # Connect Worker
       CONNECT_SECURITY_PROTOCOL: SASL_SSL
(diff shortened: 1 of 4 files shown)


ybyzek:
@imcdo please note https://confluentinc.atlassian.net/browse/DGS-72 and in light of that jira, it _should_ mean that if `CLASSPATH` is set on the connect docker image (e.g. https://github.com/confluentinc/cp-demo/blob/5.4.x/docker-compose.yml#L358) , this PR should not be required

imcdo:
> @imcdo please note https://confluentinc.atlassian.net/browse/DGS-72 and in light of that jira, it _should_ mean that if `CLASSPATH` is set on the connect docker image (e.g. https://github.com/confluentinc/cp-demo/blob/5.4.x/docker-compose.yml#L358) , this PR should not be required. So first step would be to validate if the `CLASSPATH` is set properly for the connect image in this branch.

Hey @ybyzek This is running in 5.4.5 patch release, and the classpath isn't set.  Also the class itself only exists in the schema_registry image to begin with so there would still require some change to switch over to running on the schema registry container rather than connect.
Unitl that ticket is addressed (as it seems like it hasn't been touched in a half a year or so) this is a way we can add it to the class path for this version of cp-demo for testing in the future.

ybyzek:
@imcdo how have you validated that the `MonitoringConsumerInterceptor` class doesn't exist on the connect image?  This doesn't sound right, how else does it run interceptors for its worker/connectors?

imcdo:
https://confluentinc.atlassian.net/wiki/spaces/QERM/pages/2574124059/Diggin+around+for+classpath+issues
just checked all the jars to make sure, and no jar seems to have it in it.

ybyzek:
@imcdo as discussed in Slack:

1. You verified that jar *does* exist on the Connect image
2. The workaround is not this PR.  The workaround is to modify the CLASSPATH locally in your docker-compose file to whatever is shown in the output of `docker-compose exec connect ls /usr/share/java/monitoring-interceptors/` or just change it to a wildcard like in `6.2.0-post` branch

ybyzek (changes requested):
The wildcard goes into the `CLASSPATH` parameter in docker-compose.yml, which would obviate the need for modifying the shell script.

ybyzek:
@imcdo please confirm final run-thru with latest commit and then I can merge (note to self: `pint merge` with the `--skip-commits` option so it doesn’t propagate (5.5.x already uses the wildcard: https://github.com/confluentinc/cp-demo/blob/5.5.x/docker-compose.yml#L391 so this does not need to be propagated to newer branches.)

imcdo:
> @imcdo please confirm final run-thru with latest commit and then I can merge (note to self: `pint merge` with the `--skip-commits` option so it doesn’t propagate (5.5.x already uses the wildcard: https://github.com/confluentinc/cp-demo/blob/5.5.x/docker-compose.yml#L391 so this does not need to be propagated to newer branches.)

@ybyzek looks good! validated cp-demo!  let me know if you want me to do the merge

ybyzek (approved):
@imcdo LGTM

PR #387, opened 2021-09-01
Use $REPOSITORY variable in functions.sh

imcdo:
### Description 
use the $REPOSITORY variable isntead of the hardcoded confluentinc/ for container access in functions.  This is so credential generation will work for containers not in the public confluent repository

### Author Validation

### Reviewer Tasks


ybyzek (approved):
@imcdo LGTM

PR #390, opened 2021-09-27
Devx 2672: Enable cp-demo to be run in gitpod.io

vdesabou:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2672

Allow to run cp-demo using [Gitpod.io](https://gitpod.io)

### Author Validation

- [x] Run all steps from https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/on-prem.html#module-1-on-prem-tutorial
- [x] Run all steps from https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/hybrid-cloud.html
- [ ] jmx-monitoring-stacks

### Reviewer Tasks

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,35 @@
+github:
+  prebuilds:
+    # enable for the default branch (defaults to true)
+    master: true
+    # enable for all branches in this repo (defaults to false)
+    branches: true
+    # enable for pull requests coming from this repo (defaults to true)
+    pullRequests: true
+    # add a "Review in Gitpod" button as a comment to pull requests (defaults to false)
+    addComment: true
+    # configure whether Gitpod registers itself as a status check to pull requests
+    addCheck: false
+
+tasks:
+  - init: |
+      ./scripts/start.sh && ./scripts/stop.sh
+    command: |
+      curl -L --http1.1 https://cnfl.io/ccloud-cli | sudo sh -s -- -b /usr/local/bin
+      echo "🚀 You can now follow steps in https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/overview.html"
+
+vscode:
+  extensions:
+    - ms-azuretools.vscode-docker
+
+ports:
+  - port: 1000-5600
+    onOpen: ignore
+  - port: 5601
+    onOpen: notify
+  - port: 5602-9020
+    onOpen: ignore
+  - port: 9021-9022
+    onOpen: notify
+  - port: 9023-13000
+    onOpen: ignore
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -69,7 +69,7 @@ services:
     hostname: openldap
     container_name: openldap
     ports:
-        - 389:389
+        - 1389:389
     environment:
         LDAP_ORGANISATION: "ConfluentDemo"
         LDAP_DOMAIN: "confluentdemo.io"
--- a/docs/on-prem.rst
+++ b/docs/on-prem.rst
@@ -19,6 +19,7 @@ If you prefer non-Docker examples, please go to `confluentinc/examples GitHub re
 
 After you run through the guided tutorial below, apply the concepts you learn here to build your own event streaming pipeline in |ccloud|, a fully managed, cloud-native event streaming platform powered by |ak|. When you sign up for `Confluent Cloud <https://confluent.cloud>`__, use the promo code ``CPDEMO50`` to receive an additional $50 free usage (`details <https://www.confluent.io/confluent-cloud-promo-disclaimer>`__).
 
+.. note:: Instead of running ``cp-demo`` locally with Docker, you can also run it in a cloud IDE using `gitpod.io <https://gitpod.io>`__ by following this `link <https://gitpod.io/#https://github.com/confluentinc/cp-demo>`__
 
 Prerequisites
 -------------
--- a/scripts/helper/functions.sh
+++ b/scripts/helper/functions.sh
@@ -128,8 +128,17 @@ create_certificates()
 
   # Enable Docker appuser to read files when created by a different UID
   echo -e "Setting insecure permissions on some files in ${DIR}/../security for demo purposes\n"
+  if [[ "$OSTYPE" == "darwin"* ]]
+  then
+      # Mac OS
   chmod 644 ${DIR}/../security/keypair/keypair.pem
   chmod 644 ${DIR}/../security/*.key
+  else
+      # Linux
+      sudo chmod 644 ${DIR}/../security/keypair/keypair.pem
+      sudo chmod 644 ${DIR}/../security/*.key
+  fi
+
 }
 
 build_connect_image()
--- a/scripts/ccloud/create-ccloud-workflow.sh
+++ b/scripts/ccloud/create-ccloud-workflow.sh
@@ -33,8 +33,8 @@ ccloud::generate_configs $CONFIG_FILE
 source "delta_configs/env.delta"
 
 echo
-echo "Sleep an additional 60s to wait for all Confluent Cloud metadata to propagate"
-sleep 60
+echo "Sleep an additional 90s to wait for all Confluent Cloud metadata to propagate"
+sleep 90
 
 echo -e "\nStart Confluent Replicator to Confluent Cloud:"
 CONNECTOR_SUBMITTER="User:connectorSubmitter"


docs/on-prem.rst
```diff
@@ -19,6 +19,7 @@ If you prefer non-Docker examples, please go to `confluentinc/examples GitHub re
 
 After you run through the guided tutorial below, apply the concepts you learn here to build your own event streaming pipeline in |ccloud|, a fully managed, cloud-native event streaming platform powered by |ak|. When you sign up for `Confluent Cloud <https://confluent.cloud>`__, use the promo code ``CPDEMO50`` to receive an additional $50 free usage (`details <https://www.confluent.io/confluent-cloud-promo-disclaimer>`__).
 
+.. note:: Instead of running ``cp-demo`` locally with Docker, you can also run it in a cloud IDE using `gitpod.io <https://gitpod.io>`__ by following this `link <https://gitpod.io/#https://github.com/confluentinc/cp-demo>`__
```
ybyzek, line 22: Can we move this to the `Setup` section instead?  i.e., with verbiage something like "You can run this demo locally with Docker...and do these 3 steps...or you can run this demo in a cloud IDE using `gitpod.io <https://gitpod.io>`__ by following this `link <https://gitpod.io/#https://github.com/confluentinc/cp-demo>`__" (you can wordsmith)

scripts/helper/functions.sh
```diff
@@ -128,8 +128,17 @@ create_certificates()
 
   # Enable Docker appuser to read files when created by a different UID
   echo -e "Setting insecure permissions on some files in ${DIR}/../security for demo purposes\n"
+  if [[ "$OSTYPE" == "darwin"* ]]
+  then
+      # Mac OS
```
ybyzek, line 133: Format for lines below -- needs indentation.

docker-compose.yml
```diff
@@ -69,7 +69,7 @@ services:
     hostname: openldap
     container_name: openldap
     ports:
-        - 389:389
+        - 1389:389
```
ybyzek, line 72: Can you please remind me why this port change was required?

chuck-confluent:
@ybyzek Note that to enabled prebuilds, whoever has appropriate repo permissions would have to [install the gitpod app in GitHub](https://www.gitpod.io/docs/prebuilds#on-github) and restrict its access to the cp-demo repo (and any other repo where we want prebuilds).

ybyzek:
> @ybyzek Note that to enabled prebuilds, whoever has appropriate repo permissions would have to [install the gitpod app in GitHub](https://www.gitpod.io/docs/prebuilds#on-github) and restrict its access to the cp-demo repo (and any other repo where we want prebuilds).

I did this last week and, @vdesabou   can correct me if needed, but I think there is nothing outstanding to enable prebuilds.

docker-compose.yml
chuck-confluent: Maybe to avoid interfering with folks who are running ldap locally on 389? Not sure it's in the scope of this PR

chuck-confluent (approved):
So here's a cool fact: I opened https://gitpod.io/#https://github.com/vdesabou/cp-demo/blob/devx-2672 to review this PR! I can confirm cp-demo runs correctly in the workspace!

I approve

chuck-confluent:
@mikebin @rspurgeon it looks like merge is blocked by your review. Thoughts on this? It would be great to get this out.

mikebin (approved):
Agree with the comments already provided, otherwise LGTM.

chuck-confluent:
@vdesabou looks like it's just a couple of minor tweaks and then we can get this out there! I appreciate your work on this and I look forward to seeing it live after it gets merged.

ybyzek:
> @mikebin @rspurgeon it looks like merge is blocked by your review. Thoughts on this? It would be great to get this out.
> 
> Fun fact, you can actually run this PR branch IN GITPOD with this link: https://gitpod.io/#https://github.com/vdesabou/cp-demo/blob/devx-2672 :-)

I have updated the Description of this PR with a `Pre-merge Checklist` section that needs to be completed before merge.  Normally we wouldn't have such a checklist, but this is a new workflow so just want to make sure we do the preparation upfront.

docs/on-prem.rst
ybyzek: Also, sorry if I missed it, but how does a user connect to C3 if it's running in Gitpod?

docker-compose.yml
rspurgeon: Does this port need to be exposed on the host at all?

rspurgeon:
@chuck-confluent @vdesabou I tried running this PR in gitpod without success.  I'm confused by this line:
```
init: |
      ./scripts/start.sh && ./scripts/stop.sh
```

In the terminal I can see that the resources are immediately destroyed after they are created, so i'm confused how one would use the demo.  I tried following the instructions in our docs (per the note at the end of the gitpod init script), but nothing works because the services have been destroyed.

rspurgeon:
FWIW I also verified that these changes _do not_ break a local execute of the demo on Ubuntu

chuck-confluent:
@rspurgeon There's a couple gitpod specific things to consider and it might be better to have a short conversation rather than go back in forth in comments. I'll respond to some things, but if that's not helpful then a chat or maybe a "PR huddle" might go a long way.

>I tried running this PR in gitpod without success.

What is the definition of success? The demo starts and stops successfully as far as I can tell. I can subsequently start again successfully.

> In the terminal I can see that the resources are immediately destroyed after they are created, so i'm confused how one would use the demo.

This is a gitpod thing. The `init` command will only execute on prebuilds. I find [this image](https://www.gitpod.io/docs/config-start-tasks#prebuild-and-new-workspaces) from the Gitpod docs to be helpful to understand how start tasks work. After the first prebuild (which I think happens automatically on merge if prebuilds are set up, which Yeva confirmed they are), users will enter the workspace and only the `command` command will run (echo in this case). The user would then run `./scripts/start.sh` like in the directions. We could probably change the `command`  command to go ahead and start the demo for the user automatically to save them a step if we want.

> how does a user connect to C3 if it's running in Gitpod?

Gitpod shows the active ports in the "remote explorer" and you can open web services on those ports in the browser. Here is a screen grab, and if you hover over the port, there will be an icon to open in a browser tab.

image

When I click the icon, it takes me to a new browser tab (notice the URL)

image

---

I have cp demo running in gitpod right now. Here's Kibana running in my browser from the Gitpod workspace, for example:

image

rspurgeon:
> This is a gitpod thing. The `init` command will only execute on prebuilds. I find [this image](https://www.gitpod.io/docs/config-start-tasks#prebuild-and-new-workspaces) from the Gitpod docs to be helpful to understand how start tasks work. After the first prebuild (which I think happens automatically on merge and any subsequent push if prebuilds are set up, which Yeva confirmed they are), users will enter the workspace and only the `command` command will run (echo in this case). The user would then run `./scripts/start.sh` like in the directions. We could probably change the `command` command to go ahead and start the demo for the user automatically to save them a step if we want.

In this case I don't see the value in running start && stop. If start fails, will it leave the workspace in a semi started / broken state for subsequent users or have we verified the prebuild stage fails in a graceful way? If the goal is to cache docker images, we could use `(cd scripts; source env.sh; docker-compose pull)` in the prebuild init stage instead.

Then if the goal is for the user to run the start script themselves, we should point them directly to the On-Prem instructions instead of the landing page for cp-demo (which could could be confusing if you land there from GitPod): https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/on-prem.html#cp-demo-on-prem-tutorial.  I think the instructions from here could also be confusing because the user doesn't know if they need to install pre-reqs, do the git clone and checkout, etc...   Maybe the command script shoudl have instructions added that are more specific to the GitPod experience and then point the user to a specific anchor in the cp-demo instructions where they should pickup

chuck-confluent:
How the `init` step works is it must succeed for the prebuild to be successful and the snapshot to be created. Users will never start from a failed state in the way you described. They will start from a previous successful prebuild snapshot if it exists, or in the worst case, it will run from scratch (not ideal, so we might want to ad a check whether the prebuild passes). The start/stop scripts are actually an appropriate use of `init` because it actually tests whether the demo works end to end. A docker pull would not catch new code pushes that break the demo. We don't want a prebuild to succeed if it isn't actually able to complete the demo. On the other hand, new code pushes are already tested in other CI systems, so that may be unnecessary. I could go either way.

I actually just manually triggered a successful prebuild ([doc](https://www.gitpod.io/docs/prebuilds#manual-execution-of-prebuild)) on Vincent's branch, so [this link](https://gitpod.io/#https://github.com/vdesabou/cp-demo/blob/devx-2672) should now be a little more indicative of the user experience (prebuild saved 14 minutes for the user!).

> Then if the goal is for the user to run the start script themselves, we should point them directly to the On-Prem instructions instead of the landing page for cp-demo

I agree, and I'll take back my review green light and suggest also putting `./scripts/start.sh` in the `command` start task to be run for users. I think also there should be something in the gitpod paragraph about opening browser tabs from ports in the "Remote Explorer".

chuck-confluent (changes requested):
I would request:

- add `./scripts/start.sh` to the `command` portion of the start task so that cp demo starts on launch for users.
- Point users to the on prem instructions [guided tutorial](https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/on-prem.html#guided-tutorial) rather than cp demo overview
- In the gitpod paragraph, add instructions about how to open browser tabs from open ports, preferably with screenshots

.gitpod.yml
```diff
@@ -0,0 +1,35 @@
...
+      ./scripts/start.sh && ./scripts/stop.sh
+    command: |
+      curl -L --http1.1 https://cnfl.io/ccloud-cli | sudo sh -s -- -b /usr/local/bin
+      echo "🚀 You can now follow steps in https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/overview.html"
+
+vscode:
+  extensions:
+    - ms-azuretools.vscode-docker
+
+ports:
```
chuck-confluent, line 25: I have no idea why, but I'm getting notifications for all the ports

rspurgeon:
>  actually just manually triggered a successful prebuild ([doc](https://www.gitpod.io/docs/prebuilds#manual-execution-of-prebuild)) on Vincent's branch, so [this link](https://gitpod.io/#https://github.com/vdesabou/cp-demo/blob/devx-2672) should now be a little more indicative of the user experience (prebuild saved 14 minutes for the user!).

I think it's a little confusing for the user to see the entire output of the test run (init stage) when they go to a prebuild link.   Can we either supress that w/ GitPod somehow or can we redirect output of the init stage to a file so the user gets a clean looking terminal when they initiate a workspace from the prebuilt snapshot?

scripts/helper/functions.sh
vdesabou: done in 0c86b59f94d03ab709be905d35c8d1c5acba914a

docker-compose.yml
vdesabou: I changed it because I'm getting on Gitpod permissions issues and ports used < 1024 for example, see:

CleanShot 2021-09-17 at 12 06 08
 
But I agree with @rspurgeon, there is no need to expose it, it is only used from within containers

Should I remove it completely or keep it like that ?

vdesabou:
Sorry all, I did not get notifications and completely missed your comments

Regarding @chuck-confluent comments:

> add ./scripts/start.sh to the command portion of the start task so that cp demo starts on launch for users.

Yes I thought about it but a common use case for GTS is that we make a few modifications _before_ starting cp-demo. Therefore I propose to make it optional by doing the following:

```
      if [ -z "$DISABLE_AUTOSTART" ]; then echo "🚀 Starting up cp-demo (you can disable autostart by exporting DISABLE_AUTOSTART environment variable, see https://www.gitpod.io/docs/environment-variables";./scripts/start.sh; echo "🚀 You can now follow steps in https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/on-prem.html#guided-tutorial"; else echo "ℹ️ DISABLE_AUTOSTART environment variable is set, use ./scripts/start.sh to start cp-demo";fi
```

This way if `DISABLE_AUTOSTART` is set (see [docs](https://www.gitpod.io/docs/environment-variables)), then we don't start automatically and we print:

```
ℹ️ DISABLE_AUTOSTART environment variable is set, use ./scripts/start.sh to start cp-demo
```

Otherwise we have:

```
🚀 Starting up cp-demo (you can disable autostart by exporting DISABLE_AUTOSTART environment variable, see https://www.gitpod.io/docs/environment-variables)

./scripts/start.sh ****** this is executed

🚀 You can now follow steps in https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/on-prem.html#guided-tutorial
``

> Point users to the on prem instructions guided tutorial rather than cp demo overview

Agreed, see above


> In the gitpod paragraph, add instructions about how to open browser tabs from open ports, preferably with screenshots

ok

File: .gitpod.yml (hunk 4 above)
Comment (vdesabou) [reply .gitpod.yml:25]: I do see the same, need to investigate why

File: docker-compose.yml (hunk 3 above)
Comment (rspurgeon) [reply docker-compose.yml:72]: Yes, I believe we can remove this mapping entirely as long as we verify no broken behavior (of course)

File: .gitpod.yml (hunk 4 above)
Comment (vdesabou) [reply .gitpod.yml:25]: I can reproduce same issue as reported here https://github.com/gitpod-io/gitpod/issues/1484
I'll go ahead and define ports one by one

Comment (vdesabou) [reply .gitpod.yml:25]: With the modification, I only see this now:

CleanShot 2021-10-06 at 12 12 10@2x

And same for Kibana

File: docs/on-prem.rst (hunk 1 above)
Comment (vdesabou) [reply docs/on-prem.rst:22]: I've moved it to Setup section like this:

CleanShot 2021-10-06 at 13 58 37

File: docker-compose.yml (hunk 3 above)
Comment (vdesabou) [reply docker-compose.yml]: I've removed it entirely

**Comment by chuck-confluent** [comment]
I'm sorry, this is too cool:

image

I can open C3 as a preview inside the IDE and browse code while I browse topics in C3 or dashboards in kibana. This is SO exciting!

File: docs/on-prem.rst [hunk 5]
```diff
@@ -37,6 +36,8 @@ This example has been validated with:
 Setup
 -----
 
+You can run this demo locally with Docker and follow the 3 steps below, or you can run it in a cloud IDE using `Gitpod.io <https://gitpod.io>`__ workspace by following this `link <https://gitpod.io/#https://github.com/confluentinc/cp-demo>`__.
```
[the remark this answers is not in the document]
ybyzek: `pint` will work on cp-demo post PR merge.

> I need to test, but I was hoping that since 6.2.1-post is default, it will redirect to this branch

Does this mean Gitpod will always use whatever branch is default?  (Sorry, I don't know much about Gitpod).  Default branch in the repo right now is 6.2.1-post, does that mean a user who is looking at 6.2.0 version of docs, they click on this link expecting 6.2.0 version of cp-demo, but they get 6.2.1 version of cp-demo?  This will always be wrong/unexpected for older versions...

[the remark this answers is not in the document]
chuck-confluent: I fairly sure gitpod uses the github link you put in `gitpod.io/#<github link>`, which would use the default branch.

To be explicit, you can specify the specific branch in the gitpod link, like I did earlier with https://gitpod.io/#https://github.com/vdesabou/cp-demo/blob/devx-2672 . RST might have a variable that can be used for the link? Might prefer to just use the default branch though.

PR #392, opened 2021-10-08
fix: Dockerfile-confluenthub connector uses REPOSITORY override if defined

andrewegel:
This seems to be a long standing bug that I'm fixing here.

Locally tested against ECR images with minor `env_files/config.env` changes.

### DIFF
--- a/Dockerfile-confluenthub
+++ b/Dockerfile-confluenthub
@@ -14,8 +14,8 @@
 # limitations under the License.
 
 ARG CP_VERSION
-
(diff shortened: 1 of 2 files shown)


ybyzek:
@andrewegel these files have been renamed in more recent branches, and this bug has been fixed, please see https://github.com/confluentinc/cp-demo/blob/6.2.1-post/Dockerfile#L19

andrewegel:
> @andrewegel when it comes to `pint merge` can you please limit it to this branch?

I'll work an applicable change up through the branches. Skipping where needed.

PR #395, opened 2021-10-12
Fix Gitpod C3 ksqlDB editor Connection Bug

chuck-confluent:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2679

Gitpod had a bug where the C3 advertised listener for ksqlDB wikipedia app needed to use the workspace url rather than localhost in order to run the websocket connection from c3 to ksqlDB.

### Author Validation

- [N/A] Run cp-demo in gitpod when `C3_KSQLDB_HTTPS=true` -- This option doesn't work in gitpod, so I'd like to limit the scope of this PR to not include this scenario.
- [x] Run cp-demo in gitpod when `C3_KSQLDB_HTTPS=false` 
- [x] Run cp-demo locally when `C3_KSQLDB_HTTPS=true` (at least ensure proper branch execution)
- [x] Run cp-demo locally when `C3_KSQLDB_HTTPS=false`  (at least ensure proper branch execution)

I verified that ksqlDB queries now work through the c3 ksqldb query editor.

### Reviewer Tasks

### DIFF
--- a/scripts/security/ldap_users/20_group_add.ldif
+++ b/scripts/security/ldap_users/20_group_add.ldif
@@ -71,7 +71,7 @@ memberuid: cn=appSA,ou=users,{{ LDAP_BASE_DN }}
 dn: cn=KafkaDevelopers,ou=groups,{{ LDAP_BASE_DN }}
 changetype: modify
 add: memberuid
-memberuid: cn=badapp,ou=users,{{ LDAP_BASE_DN }}
+memberuid: cn=badApp,ou=users,{{ LDAP_BASE_DN }}
 
 dn: cn=KafkaDevelopers,ou=groups,{{ LDAP_BASE_DN }}
 changetype: modify


chuck-alt-delete:
Yeah that https setting doesn’t work. Can’t seem to open control center from 9022-<GitPod url> even though I can curl -k localhost:9022 from the workspace

chuck-alt-delete:
The ksqlDB port must be public so that the c3_ksqldb advertised listeners setting will work. The websocket connection for the ksqlDB editor needs to be able to reach ksqlDB via advertised listener from the browser. See https://docs.confluent.io/platform/current/ksqldb/integrate-ksql-with-confluent-control-center.html#network-configuration-for-ksqldb-and-c3-short , bullet 3.

I also made the c3 port public so that users can share that link with friends and they can looks at c3 together. Essentially “why not?” for that port.

vdesabou:
I guess an action for me would be to add to the documentation that C3_KSQLDB_HTTPS=true is not supported with Gitpod.io, right ?

javabrett:
@chuck-alt-delete agreed - my comment on lack of GitPod experience ... if these are ephemeral as you'd expect they's have to be, there's little value in securing them.

chuck-alt-delete:
Who should I tag to merge this?

PR #398, opened 2021-10-14
fix: Fix CUB_CLASSPATH & Allow REPOSITORY override for building connector

andrewegel:
Related to an internal bug report. There are two fixes here:

* Appending a ubi8/deb9 class path location of where to find the utility-belt classes.
* Override REPOSITORY location when building the connect docker image.

My plan is to merge this through the 5.5.x post branches into 5.5.x, and then merge w/o changes into 6.0.x and above.

### DIFF
--- a/Dockerfile-confluenthub
+++ b/Dockerfile-confluenthub
@@ -14,8 +14,9 @@
 # limitations under the License.
 
 ARG CP_VERSION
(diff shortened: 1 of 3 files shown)


ybyzek:
@andrewegel are we keeping this fix to 5.5.x branch (not propagating to 6.x)?

PR #400, opened 2021-10-21
Adding .gitpod.Dockerfile based on gitpod/workspace-base

vdesabou:
### Description 

Added a custom docker image https://www.gitpod.io/docs/config-docker based on [gitpod/workspace-base ](https://github.com/gitpod-io/workspace-images/blob/master/base/Dockerfile) instead of [gitpod/workspace-full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile) in order have a small base image (we only need Docker and not all Node.js, Java, Python, etc...deps) and improve workspace startup time.

### Author Validation

- [ ] Run cp-demo
- [ ] Validate it improve performances

### Reviewer Tasks


chuck-confluent:
I'm going to close this

PR #401, opened 2021-11-02
Include JMX Monitoring stacks on Gitpod configuration

jeqo:
### Description 

While enabling gitpod on jmx-monitoring-stacks repo (https://github.com/confluentinc/jmx-monitoring-stacks/pull/54), it made sense to extend cp-demo instead of copy/paste'ing most of the gitpod configuration in another project.

This PR adds a new variable `MONITORING_STACK` to choose which stack to start, if any, when starting cp-demo on gitpod.

### Author Validation

### Reviewer Tasks

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -14,7 +14,7 @@ github:
 tasks:
   - name: prebuild
     init: |
-      cd .. && git clone https://github.com/confluentinc/jmx-monitoring-stacks && cd ../cp-demo/. \
+      cd .. && git clone https://github.com/confluentinc/jmx-monitoring-stacks && cd ../cp-demo/. && \
         ./scripts/start.sh && ./scripts/stop.sh && \
         gp sync-done prebuild
     command: |
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,6 +6,8 @@ github:
     branches: true
     # enable for pull requests coming from this repo (defaults to true)
     pullRequests: true
+    # enable for pull requests coming from forks (defaults to false)
+    pullRequestsFromForks: true
     # add a "Review in Gitpod" button as a comment to pull requests (defaults to false)
     addComment: true
     # configure whether Gitpod registers itself as a status check to pull requests
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -15,11 +15,8 @@ github:
 
 tasks:
   - name: prebuild
-    init: |
-      ./scripts/start.sh && ./scripts/stop.sh && gp sync-done prebuild
-    command: |
-      curl -L --http1.1 https://cnfl.io/ccloud-cli | \
-        sudo sh -s -- -b /usr/local/bin; exit
+    init: ./scripts/start.sh && ./scripts/stop.sh && gp sync-done prebuild
+    command: curl -L --http1.1 https://cnfl.io/ccloud-cli | sudo sh -s -- -b /usr/local/bin; exit
 
   - name: cp-demo
     init: gp sync-await prebuild
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -14,7 +14,7 @@ github:
 tasks:
   - name: prebuild
     init: |
-      cd .. && git clone https://github.com/confluentinc/jmx-monitoring-stacks && cd ../cp-demo/. && \
+      cd .. && git clone https://github.com/confluentinc/jmx-monitoring-stacks && cd cp-demo/. && \
         ./scripts/start.sh && ./scripts/stop.sh && \
         gp sync-done prebuild
     command: |
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -26,7 +26,7 @@ tasks:
     command: |
       if [ -z "$DISABLE_AUTOSTART" ]; then
         echo "🚀 Starting up cp-demo (you can disable autostart by exporting DISABLE_AUTOSTART environment variable, see https://www.gitpod.io/docs/environment-variables)";
-        if [ -z "$MONITORING_STACK"]; then
+        if [ -z "$MONITORING_STACK" ]; then
           echo "cp-demo is not including any JMX monitoring stack (you can enable them by exporting MONITORING_STACK environment variable, see https://github.com/confluentinc/jmx-monitoring-stacks)."
           ./scripts/start.sh;
         else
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -13,12 +13,31 @@ github:
 
 tasks:
   - name: prebuild
-    init: ./scripts/start.sh && ./scripts/stop.sh && gp sync-done prebuild
-    command: curl -L --http1.1 https://cnfl.io/ccloud-cli | sudo sh -s -- -b /usr/local/bin; exit
+    init: |
+      cd .. && git clone https://github.com/confluentinc/jmx-monitoring-stacks && cd cp-demo/. && \
+        ./scripts/start.sh && ./scripts/stop.sh && \
```
jeqo, line 18: @vdesabou @chuck-confluent any reason to start/stop the whole environment before starting?

vdesabou: The `init` is done as part of https://www.gitpod.io/docs/prebuilds, so it doesn't really matter if it takes more time

Doing start/stop is the easiest way to pull and build connect image + generating certificates

vdesabou:
Am I the only one getting for https://gitpod.io/#https://github.com/confluentinc/cp-demo/pull/401 ?

CleanShot 2021-11-03 at 11 22 37@2x

.gitpod.yml
jeqo: got it, thanks @vdesabou

jeqo:
@vdesabou it works for me: https://gray-bee-8m5ifxwq.ws-us17.gitpod.io/
Screenshot 2021-11-03 at 05 24 45

.gitpod.yml
jeqo: Thanks @vdesabou, that make sense.

So, every new environment will run this as the workspace is created (~15 min):
image

As start.sh does much more than pulling/building images (e.g. lot of waiting, deployment), having an `init.sh` that _just_ pull/build images and generate certs could improve user experience (maybe reduce it to ~5min). wdyt? If worth, I can create another PR for it.

jeqo: aha! So, I might be hit by this: 

```
    # enable for pull requests coming from forks (defaults to false)
    pullRequestsFromForks: true
```

Should we enable that one for forks as well? Seems we are enabling on most cases anyway.

vdesabou: Yes I wasn't sure to put it or not, but I think it makes sense to allow it

jeqo: adding it

.gitpod.yml
```diff
@@ -6,19 +6,40 @@ github:
...
     addComment: true
     # configure whether Gitpod registers itself as a status check to pull requests
     addCheck: false
 
 tasks:
   - name: prebuild
-    init: ./scripts/start.sh && ./scripts/stop.sh && gp sync-done prebuild
-    command: curl -L --http1.1 https://cnfl.io/ccloud-cli | sudo sh -s -- -b /usr/local/bin; exit
+    init: |
+      cd .. && git clone https://github.com/confluentinc/jmx-monitoring-stacks && cd cp-demo/. && \
```
chuck-confluent, line 19: I think the jmx repo should only be cloned if the user specifies the MONITORING_STACK variable. I think this should move to the "cp-demo" task

.gitpod.yml
```diff
@@ -6,19 +6,40 @@ github:
...
+        ./scripts/start.sh && ./scripts/stop.sh && \
```
chuck-confluent, line 20: Something I'm planning to do anyway is to replace `./scripts/start.sh && ./scripts/stop.sh` with simply `CLEAN=true ./scripts/start.sh`. This makes a better experience if there is no prebuild enabled. There's no reason to require stop.sh. I think it's fine to add it to this PR because I don't want you to have to rebase and fix the conflict.

.gitpod.yml
```diff
@@ -6,19 +6,40 @@ github:
...
 
   - name: cp-demo
     init: gp sync-await prebuild
-    command: if [ -z "$DISABLE_AUTOSTART" ]; then echo "🚀 Starting up cp-demo (you can disable autostart by exporting DISABLE_AUTOSTART environment variable, see https://www.gitpod.io/docs/environment-variables)";./scripts/start.sh; echo "🚀 You can now follow steps in https://docs.confluent.io/platform/current/tutorials/cp-demo/docs/on-prem.html#guided-tutorial"; else echo "ℹ️ DISABLE_AUTOSTART environment variable is set, use ./scripts/start.sh to start cp-demo";fi
+    command: |
+      if [ -z "$DISABLE_AUTOSTART" ]; then
+        echo "🚀 Starting up cp-demo (you can disable autostart by exporting DISABLE_AUTOSTART environment variable, see https://www.gitpod.io/docs/environment-variables)";
+        if [ -z "$MONITORING_STACK" ]; then
+          echo "cp-demo is not including any JMX monitoring stack (you can enable them by exporting MONITORING_STACK environment variable, see https://github.com/confluentinc/jmx-monitoring-stacks)."
+          ./scripts/start.sh;
```
chuck-confluent, line 33: change to `CLEAN=true ./scripts/start.sh`

.gitpod.yml
```diff
@@ -6,19 +6,40 @@ github:
...
+        else
+          echo "cp-demo includes JMX monitoring stack $MONITORING_STACK"
+          cd ../jmx-monitoring-stacks
+          ./$MONITORING_STACK/start.sh;
```
chuck-confluent, line 37: I would suggest the JMX repo also set `CLEAN=true` when starting CP Demo in its own start scripts. My thinking is it's possible for a prebuild to create certificates and junk that should probably be regenerated.

chuck-confluent (changes requested):
@jeqo Hey, I'm taking over maintenance of CP demo and I'd like to get this PR merged. Just a couple of minor comments, but mostly looks good!

chuck-confluent:
I'm working on some other gitpod prebuild optimizations in https://github.com/confluentinc/cp-demo/pull/419 that would need to go through before adding this. Hold tight!

chuck-confluent:
Sorry for the delay on this. I've since discovered that CP demo is pushing the resource limitations of gitpod quite a bit, so i don't think it's feasible to run more components. I'll have to close this PR, unfortunately.

PR #402, opened 2021-11-03
DEVX-2686: make null-safe reference to GITPOD_WORKSPACE_URL

javabrett:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2686

Fix this error when running `add-broker.sh` from SBC demo:

```
cp-demo/scripts/sbc/../env.sh: line 30: GITPOD_WORKSPACE_URL: unbound variable
```

### Author Validation

 - [ ] Documentation
 - [X] Run cp-demo 

### Reviewer Tasks

### DIFF
--- a/scripts/env.sh
+++ b/scripts/env.sh
@@ -27,7 +27,7 @@ else
 fi
 
 # Gitpod only supports the C3_KSQLDB_HTTPS=false scenario and exposes services with a custom URL
-if [[ -n "$GITPOD_WORKSPACE_URL" ]]; then
+if [[ -n "${GITPOD_WORKSPACE_URL:-}" ]]; then
   C3_KSQLDB_HTTPS="false"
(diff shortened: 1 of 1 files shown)


ybyzek (commented):
@javabrett thanks for triaging and the quick turnaround.  I haven't validated the fix but based on what I read up on the error message, it seems to be a great fix.  

1. Please retarget and rebase to `6.2.0-post`, which is when GitPod was introduced into cp-demo
2. It would be good to get a second review from @chuck-confluent before merging.

javabrett:
@ybyzek target branch changed to `6.2.0-post`.

javabrett:
Tested local on MacOS and in GitPod, running `add-broker.sh`.

ybyzek (approved):
From Slack, sounds like @esakkiraj has validated this PR in his env, so we don't have to wait for @chuck-confluent .  @javabrett please merge at your convenience.

PR #408, opened 2021-12-02
Add docker-utils.jar for all CUB_CLASSPATH

andrewegel:
Minor fix.


ybyzek (approved):
Approving (but without context)

PR #418, opened 2022-04-07
Replace docker-compose exec with docker exec

chuck-confluent:
### Description 

https://confluentinc.atlassian.net/browse/TMM-76

There is a breaking change in the TTY output behavior of `docker-compose exec` between docker compose v1 and v2:
- https://github.com/docker/compose/issues/9202

Since we explicitly set `container_name` in `docker-compose.yml`, `docker exec` works the same as `docker-compose exec`, except by not using `docker-compose`, we avoid the discrepancy between dc v1 and v2.

### Author Validation

- [x] Run cp-demo

### Reviewer Tasks

- [ ] Run cp-demo

### DIFF
--- a/docs/hybrid-cloud.rst
+++ b/docs/hybrid-cloud.rst
@@ -239,8 +239,8 @@ Metrics API
 
    .. code-block:: text
 
-      CURRENT_TIME_MINUS_1HR=$(docker-compose exec --no-TTY tools date -Is -d '-1 hour' | tr -d '\r')
-      CURRENT_TIME_PLUS_1HR=$(docker-compose exec --no-TTY tools date -Is -d '+1 hour' | tr -d '\r')
+      CURRENT_TIME_MINUS_1HR=$(docker exec tools date -Is -d '-1 hour' | tr -d '\r')
+      CURRENT_TIME_PLUS_1HR=$(docker exec tools date -Is -d '+1 hour' | tr -d '\r')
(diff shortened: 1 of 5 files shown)


chuck-confluent:
Running locally with docker compose v 1.29. Will update.

Schm1tz1:
Thanks for the PR, that saved my day. I ran into issues after updating cp-demo on MacOS with most recent versions of docker and compose available.

CP successfully up and running on my system:
- MacOS 12.3.1
- docker-compose version 1.29.2
- Docker engine version 20.10.14

chuck-confluent:
@Schm1tz1 btw, the latest version of docker-compose is 2.4, not 1.29. It's a whole thing. It's what ended up breaking cp demo. We should get this PR merged soon and it will work on both 2.4 and 1.29.

Schm1tz1:
@chuck-confluent you are absolutely correct. The problem is, you only get that version if you install it e.g. with homebrew or pip. Compose 1.29 still comes bundled with docker-desktop and that will affect many users like me before manually installing the newer compose.

chuck-confluent:
@Schm1tz1 I think the next docker for desktop update will update compose to 2.4. I'm already seeing folks on mac like this person: https://github.com/docker/compose/issues/9202#issuecomment-1092777992

jwfbean (approved):

PR #419, opened 2022-04-07
modify gitpod init for better behavior when prebuilds are not available

chuck-confluent:
### Description 

Right now, the `init` step in gitpod starts and stops cp demo. If a prebuild isn't available, that means users will experience start, stop, and start again.

Other improvements that may be in scope:
- limit the number of prebuilds that run to only the default branch and PRs.
- add prebuilds as a status check to PRs to act as some light CI / automated validation 
- fix transient issue where zookeeper gets confused after restart and brokers barf
- fix issue where connect isn't ready in time (more of a hack, but a proper fix is out of scope)

### Author Validation

- [ ] Documentation
- [x] Run cp-demo

### Reviewer Tasks

- [ ] Documentation
- [ ] Run cp-demo

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,13 +3,13 @@ github:
     # enable for the default branch (defaults to true)
     master: true
     # enable for all branches in this repo (defaults to false)
-    branches: true
+    branches: false
     # enable for pull requests coming from this repo (defaults to true)
     pullRequests: true
     # add a "Review in Gitpod" button as a comment to pull requests (defaults to false)
     addComment: true
     # configure whether Gitpod registers itself as a status check to pull requests
-    addCheck: false
+    addCheck: true
 
 tasks:
   - name: prebuild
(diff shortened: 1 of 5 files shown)


chuck-confluent:
I just want to run through this one more time to verify what behavior looks like after successful prebuild

chuck-confluent:
Gitpod prebuilds have been really flakey. It’s hard to tell if it’s gitpod or us. I’m any case, the goal of this PR is achieved I think.

chuck-confluent:
Did some timed tests just to see:

Gitpod:
- prebuild: ~6.5 minutes from click to end of start.sh
- no prebuild: ~10 minutes

Local (tested on lenovo x1 32 GB RAM, hex core Intel(R) Core(TM) i7-8850H CPU running windows 11 and WSL2 ubuntu 18.04 with roughly 100 Mbps download speed):
- For a new user or new version of CP ~14 minute runtime of start.sh
- 2nd run: ~3.5 minutes

One big advantage with gitpod is the workspaces are running in GCP with very high network bandwidth, so the pulls are faster, and loading a prebuild is faster than downloading the docker images from the internet.

jwfbean (approved):

PR #420, opened 2022-04-12
TMM-29: Add Health+ Configuration Placeholders in Comments

chuck-confluent:
### Description 

https://confluentinc.atlassian.net/jira/software/c/projects/TMM/issues/TMM-29

This PR adds comments in docker-compose.yml about the configurations required for Confluent Health+

### Author Validation

These are just comments, so they don't affect runtime behavior

### Reviewer Tasks

Is it readable? Is it correct?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -158,6 +158,16 @@ services:
       KAFKA_AUTO_CREATE_TOPICS_ENABLE: 'false'
       KAFKA_DEFAULT_REPLICATION_FACTOR: 2
 
+      # Confluent Health+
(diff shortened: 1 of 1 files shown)


chuck-confluent:
I should mention that I consider this to be separate from adding on a health+ sign-up tutorial. We can choose that it's not ok to comment about health+ without an accompanying tutorial, in which case we can extend this PR to include https://confluentinc.atlassian.net/browse/TMM-32 (revamp CP Demo to use Health+ rather than C3 monitoring)

mkananizadeh (approved):

PR #428, opened 2022-08-08
add descriptions and reorder ports for gitpod

chuck-confluent:
Quality of life update for gitpod:
- names and descriptions for ports in gitpod port explorer


chuck-confluent:
Looks like the ordering isn't taken into account. Still, it's nice to have some descriptions rather than just port numbers.

image

PR #453, opened 2024-02-05
feat(): enable confluent.consumer.lag.emitter

pneff93:
### Description 

Since CP 7.5, we can emit consumer lag via the configuration confluent.consumer.lag.emitter.enabled, see https://docs.confluent.io/platform/current/monitor/monitor-consumer-lag.html#monitor-consumer-lag

To use this functionality in the [jmx-monitoring-stacks](https://github.com/confluentinc/jmx-monitoring-stacks) repository, we first need to enable it in cp-demo.

### Author Validation

### Reviewer Tasks


javabrett (approved):
LGTM

PR #479, opened 2025-06-30
Fixed SR registration issue in C3 - need default SR instance

javabrett:
### Description 

There has been a change around the 7.9.0 timeframe to the processing of Control Center Schema Registry registrations, with the requirement now seeming to be that a default registration is required.  Currently we only have an infixed instance `SR1`.

Since we only have one registration, we will change this to the default.  This will resolve exceptions we see in C3 logs trying to connect to `localhost:8081`.

### Author Validation

- [ ] Documentation
- [X] Run cp-demo
- [ ] jmx-monitoring-stacks

### Reviewer Tasks


kylek-confluent (approved):
LGTM

